PR #40, opened 2015-05-03
Locked divise to 2.2.1 and added test-unit gem

DJG86g:
This was needed to get the rails console running.
This also may have fixed issue with ruby 2.2.2 update.


DJG86g:
see #42

PR #53, opened 2015-12-02
Person form accepts phones and emails

corincerami:
This sets both `phones` and `emails` as a `has_many` relationship on `Person`. This allows `Person` to accept nested attributes in the new person form. Phone and Email are now STI types of Channels, since they both already had their own subclasses.

This PR also adds strong params.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -1,8 +1,8 @@
 class PeopleController < ApplicationController
   before_filter :authenticate_user!
   before_filter :set_return_path
-  load_and_authorize_resource
-  
+  authorize_resource
+
   def signin
     #This is the sign-in sheet, not anything about authentication
     @people = Person.active.all
@@ -87,8 +87,8 @@ class PeopleController < ApplicationController
   def new
     @page_title = "New Person"
     @person = Person.new(status: cookies[:status], state: 'MA')
-    @person.channels.build (attributes = {category: 'E-Mail', status: 'OK', usage: '1-All'})
-    @mobile = @person.channels.build (attributes = {category: 'Mobile Phone', status: 'OK', usage: '1-All'})
+    @emails = Array(@person.emails.build(category: 'E-Mail', status: 'OK', usage: '1-All'))
+    @phones = Array(@person.phones.build(category: "Mobile Phone", status: "OK", usage: "1-All"))
     respond_to do |format|
       format.html # new.html.erb
       format.json { render json: @person }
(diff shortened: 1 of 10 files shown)


corincerami:
@kgf This PR is ready for review.

Gemfile
```diff
@@ -29,6 +30,7 @@ group :test, :development do
   gem 'database_cleaner', '1.0.1'
   gem 'test-unit'
   gem 'poltergeist'
+  gem 'pry'
```
kgf: Not for this PR, but pry-nav might be a good (better? ) addition.

marnen: Or byebug.

app/models/person.rb
```diff
@@ -1,17 +1,16 @@
...
+  include ActiveModel::ForbiddenAttributesProtection
   before_save :title_order
 
-  attr_accessible :firstname, :lastname, :status, :icsid, :department, :city, :state, :zipcode, :start_date, :end_date , :title, :gender, :date_of_birth,:division1, :division2, :channels_attributes, :title_ids, :title_order, :comments
-
-  #Having a condition on this association allows all the chaining magic to happen.
-  #Could I use a named scope, and/or could I have another association for 'active_certs' ?
   has_many :certs, :conditions => {:status =>'Active' }
-
-  # Since phones and emails derive off of channels, is "has_many :channels" below redundant?
```
kgf, line 12: Chris, you deleted the comment/question, and left the `has_many` call. Do you know the answer ?

app/models/person.rb
```diff
@@ -23,7 +22,6 @@ class Person < ActiveRecord::Base
 
   validates_presence_of :firstname, :lastname, :status
   validates_uniqueness_of :icsid, :allow_nil => true, :allow_blank => true   # this needs to be scoped to active members, or more sophisticated rules
-  validates :firstname, :uniqueness => { :scope => :lastname }
```
kgf, line 26: Why was this removed ?

app/controllers/people_controller.rb
```diff
@@ -88,6 +88,8 @@ def new
     @page_title = "New Person"
     @person = Person.new(status: cookies[:status], state: 'MA')
     @person.channels.build (attributes = {category: 'E-Mail', status: 'OK', usage: '1-All'})
+    @person.phones.build
+    @person.emails.build
     @mobile = @person.channels.build (attributes = {category: 'Mobile Phone', status: 'OK', usage: '1-All'})
```
kgf: I would think we don't need both this and line 92. These lines are to build objects to be populated from the form. Also, these should only be on the new action, not the edit.

app/controllers/people_controller.rb
```diff
@@ -102,8 +104,8 @@ def edit
   end
```
kgf, line 103: On line 103, we should make sure that phones and emails are handed into the form. I believe we are broken right now.

app/models/person.rb
corincerami, line 26: What if you have 2 `John Smith`s?

app/models/person.rb
corincerami, line 12: has_many channels is necessary if you ever want to check all of a `person`'s channels, without thinking about whether they are `phone`s or `email`s or whatnot. `person.channels` will give you all of the means of contacting a person. I'll let you decide if you think that's valuable or not.

kgf, line 12: Thanks. I intend to keep it.

db/schema.rb
```diff
@@ -189,23 +188,6 @@
     t.datetime "updated_at",     :null => false
   end
 
-  create_table "notifications", :force => true do |t|
```
kgf, line 192: I'm confused by these changes to schema. It looks like you are removing multiple tables that deal with notifying people ?

corincerami, line 192: Hm, I'm not sure where those schema changes came from. I'll fix the PR.

corincerami:
@kgf I've fixed the schema.rb

houhoulis:
I'm sorta guessing the PR doesn't need a change any more  :smiley_cat:

houhoulis:
thanks @chrisccerami !

Issue #54, opened 2015-12-10
Convert all tests to RSpec


marnen:
As discussed with @kgf , I'm going to attempt to convert the whole test suite to RSpec.

marnen:
- [x] Upgrade to RSpec 3
- Convert Test::Unit to RSpec
  - [x] unit
  - [ ] functional (replace with Cucumber?)
- [ ] Refactor specs

marnen:
For the RSpec 3 upgrade, I'm using the procedure documented at https://relishapp.com/rspec/docs/upgrade.

kgf:
Thanks for your help @marnen !

PR #57, opened 2015-12-10
[#54] Upgrade to RSpec 3 and convert unit tests

marnen:
See #54. I've upgraded to RSpec 3 and got all unit tests converted from Test::Unit to RSpec. We still need to convert functional tests, but that will be a separate pull request.


spec/support/shoulda-matchers.rb
```diff
@@ -0,0 +1 @@
+require 'shoulda/matchers'
```
kgf, line 1: Did you use any of these ?

marnen, line 1: Yes. See person_spec, for example.

PR #62, opened 2016-01-14
52 add inventory image upload

DJG86g:


app/views/items/_form.html.erb
```diff
@@ -1,10 +1,12 @@
-<%= simple_form_for(@item) do |f| %>
+<%= simple_form_for (@item), :html => {:multipart => true} do |f| %>
```
houhoulis, line 1: Ruby can have trouble if there's a space between a method call and the parentheses around its argument(s).  Unless you are writing it the exact way you did for a reason, I'd recommend using either parentheses _or_ space but not both.  Either of these should work:

``` ruby
<%= simple_form_for(@item, :html => {:multipart => true}) do |f| %>
```

(I'm guessing that "@item" and ":html => ..." are both parameters to the `simple_form_for` method.  If so, they should both be inside the parentheses if you use them.)

kgf:
This PR was made obsolete, as the code is in a different topic branch. That topic branch will be merged in soon.

PR #68, opened 2016-01-18
64 add inspections

kgf:
A WIP, but I don't want this to be a long running feature branch !


kgf:
Working in getting to green now.

app/views/items/_form.html.erb
```diff
@@ -1,8 +1,8 @@
-<%= simple_form_for(@item) do |f| %>
+<%= simple_form_for (@item), :html => {:multipart => true}  do |f| %>
```
houhoulis: I haven't tested this, but the syntax seems problematic.  Could you remove the space after `simple_form_for` and change the params?

I am not familiar with `simple_form_for`.  I apologize if this is how `simple_form_for` works.  If it's not something special about `simple_form_for`, then let me explain where I'm coming from.  I said this on a different PR:

> Ruby can have trouble if there's a space between a method call and the parentheses around its argument(s). Unless you are writing it the exact way you did for a reason, I'd recommend using either parentheses or space but not both. Either of these should work:

``` ruby
      <%= simple_form_for(@item, :html => {:multipart => true}) do |f| %>
```

> (I'm guessing that "@item" and ":html => ..." are both parameters to the simple_form_for method. If so, they should both be inside the parentheses if you use them.)

If this change was made by a generator, let me know what generator -- I am curious!

houhoulis:
+100 for avoiding a long-running feature branch!  :smiley_cat:  I am still concerned that this PR is so large, though.  It certainly includes things other than "Add inspections."

I have some ideas about the specifics.  I'll follow up with you directly.

kgf:
It was a true sprint to be ready for the demo, thus larger than I would have liked.

app/views/items/index.html.erb
```diff
@@ -17,18 +17,21 @@
...
+      	<td><%= item.location %></td>
       	<td><%= item.category %></td>
       	<td><%= item.serial1 %></td>
       	<td><%= item.serial2 %></td>
         <td><%= item.serial3 %></td>
-        <td><%= item.person.fullname unless item.person_id.blank? %></td>
-      	<td><%= item.status %></td>
+        <td><%= number_to_currency item.recent_costs %></td>
+      	<td>
+          <span class="label <%= 'label-warning' if item.status == 'Out of Service' %><%= 'label-success' if item.status == 'Available' %>">
```
houhoulis: Could you shorten this line?  One approach would be to move some of this logic into the controller.

app/views/items/show.html.erb
```diff
@@ -1,16 +1,21 @@
 <% content_for :sidebar do %>
   <%= link_to 'Return to Listing', items_path, :class => 'btn' %></br></br>
   <%= link_to 'Edit Item', edit_item_path(@item), :class => 'btn btn-primary' unless cannot? :edit, @item  %></br></br>
+  <%= link_to 'Add Repair', new_item_repair_path(@item), :class => 'btn btn-primary' unless cannot? :edit, @item  %></br></br>
+  <%= link_to 'Add Inspection', new_item_inspection_path(@item), :class => 'btn btn-primary' unless cannot? :edit, @item  %></br></br>
```
houhoulis: Could you also shorten these 2 lines, and the long lines in `spec/features/person_edit_spec.rb`?

houhoulis:
I have sorta broad questions about some of the naming and the schema changes/migrations.  I'm concerned about some details, such as adding no-content specs, and about the large breadth of this PR.

It's always good to land work rather than getting into a long-running feature branch, though.  I've reached out to @kgf to get clarity.

app/views/items/_form.html.erb
kgf: I have made a change, will push with passing tests.

kgf:
I have addressed some of your concerns, and I'm down to 4 failing tests.

houhoulis:
I have split this PR up and resubmitted it as PRs #72 , #73 , #74 , #75, #76 , and (possibly) #77 .  Assigned to @kgf.

kgf:
This has been broken up by @houhoulis into more manageable sized commits.

Thanks !

Issue #64, opened 2016-01-14
Add inspection model


kgf:
Currently inspections are intertwined with repairs, but I feel this is incorrect. Inspections are their own object. Also, the UI for inspections should be mobile first.

kgf:
Resolved by #72

PR #72, opened 2016-01-31
Add (re-add) Inspections. They belong to Item now.

houhoulis:
- Change logic to belong to Item rather than Person.
- Add some partials, view logic, and scaffolding.
- Add migration to drop & recreate table.
- Remove previous migration creating the table.


houhoulis:
Addresses #64 .
Candidate to replace #68.

app/models/inspection.rb
```diff
@@ -1,6 +1,7 @@
 class Inspection < ActiveRecord::Base
-  attr_accessible :inspection_date, :person_id, :status
-  
-  belongs_to :person
+
+  belongs_to :item
+
+  STATUS_CHOICES = ['Complete - Passed', 'Complete - Failed', 'Incomplete']
```
kgf, line 5: In looking at this, I realize this list is incomplete, as you could have "Incomplete - Failed". (If the engine is seized, don't bother checking the turn signals on a vehicle) 
Seeing the new list, it suggests that there are really two concepts here, completion, and results. You just can't have "Incomplete - Passed". Something for us to consider for a later enhancement.

spec/factories.rb
```diff
@@ -45,7 +46,7 @@
   end
   factory :inspection do
     status "Passed"
```
kgf, line 48: This also shows that status is inadequate. In fact, this is an invalid status, not one from the list.

PR #73, opened 2016-01-31
Incl MIGRATIONS: Add Question & InspectionQuestion

houhoulis:


houhoulis:
I am submitting a few PRs that I chained together while working on them.  So, the list of commits will get longer and longer (repeats).
I'll definitely rebase the PRs separately if needed.

app/views/inspections/index.html.erb
```diff
@@ -11,8 +9,6 @@
 
 <% @inspections.each do |inspection| %>
   <tr>
-    <td><%= inspection.person_id %></td>
```
kgf, line 14: It seems like this does not leave any info about the inspection in the listing, only the buttons. That said, we rarely care about a listing of inspections outside the context of an item.

PR #74, opened 2016-01-31
Add cost to repairs

houhoulis:
Add `cost` to Repair (MIGRATION), and associate Repair with Item.


app/models/repair.rb
```diff
@@ -1,8 +1,15 @@
...
+                  :status, :user_id, :cost
   belongs_to :item
   belongs_to :person
-  
+  validates_numericality_of :cost
+
   STATUS_CHOICES = ['Needed', 'In-progress', 'Awaiting Parts', 'Completed - Repaired', 'Completed - No Trouble Found']
   CATEGORY_CHOICES = ['Repair', 'Inspection' ]
+
+  def repairer_name
```
kgf, line 12: I don't think simply tying to people is going to work. Sometimes repairs will be made in house, and that will work, but  other times we'll send it out to a garage. The garages that are used are usually a very small set.

PR #76, opened 2016-01-31
Fix specs and general housecleaning

houhoulis:
This PR fixes the last few broken specs.  It also contains:
- leftover bits from the original PR I was breaking up
- some small fixes (typos & omissions) and whitespace editing.
- a commit that removes the `FactoryGirl.` prefix from every file in the `spec/` folder
- a commit to move all spec factories into their own files rather than the conglomerate `spec/factories.rb` file.
- a commit adding the `pry` gem.

It's the end of the chained PRs I have today.  It's a lot of change -- 10 commits -- and I will be happy to break them up into smaller PRs.


app/controllers/people_controller.rb
```diff
@@ -87,8 +88,8 @@ def show
   def new
     @page_title = "New Person"
     @person = Person.new(status: cookies[:status], state: 'MA')
-    @emails = Array(@person.emails.build(category: 'E-Mail', status: 'OK', usage: '1-All'))
-    @phones = Array(@person.phones.build(category: "Mobile Phone", status: "OK", usage: "1-All"))
+    @person.emails.build(category: 'E-Mail', status: 'OK', usage: '1-All')
+    @person.phones.build(category: "Mobile Phone", status: "OK", usage: "1-All")
```
kgf, line 92: I like this pattern better.

PR #77, opened 2016-01-31
Make item location not point to a location

houhoulis:
I do not think this code is desirable, but I am saving it here in PR form for later discussion.

It's chained on top of #75, just because it was easiest for me.  I'm happy to rebase if needed... if we want to go in the direction of Item.location not pointing to a Location.


kgf:
We _are_ going to have an item always point to an instance of a location object. The compromise is that there should be an instance that is "ROW (Rest OF World)" for when you don't care/know where precisely it is.

PR #81, opened 2016-02-04
schema.rb has changed. (Original migrations gone?)

houhoulis:
Running migrations produces these changes.

_[Edit:]_  This PR changes `schema.rb` by removing
three tables and altering a fourth table.  However, I
can't find any previous migrations in the repo that
would have created the tables as they are before
this PR.

### DIFF
--- a/db/migrate/20121020093514_add_status_to_inspections.rb
+++ b/db/migrate/20121020093514_add_status_to_inspections.rb
@@ -1,5 +1,0 @@
-class AddStatusToInspections < ActiveRecord::Migration
-  def change
-      add_column :inspections, :status, :string
-  end
-end
--- a/db/migrate/20130516195515_addsourcetoitem.rb
+++ b/db/migrate/20130516195515_addsourcetoitem.rb
@@ -4,7 +4,6 @@ class Addsourcetoitem < ActiveRecord::Migration
     add_column :items, :grantstart, :date
     add_column :items, :grantexpiration, :date
     add_index "certs", ["person_id"], :name => "index_certs_on_person_id"
-    add_index "inspections", ["person_id"], :name => "index_inspections_on_person_id"
   end
 
 end
--- a/db/migrate/20130820214554_correct_field_names_to_time_not_date.rb
+++ b/db/migrate/20130820214554_correct_field_names_to_time_not_date.rb
@@ -2,12 +2,10 @@ class CorrectFieldNamesToTimeNotDate < ActiveRecord::Migration
   def up
     rename_column :events, :start_date, :start_time
     rename_column :events, :end_date, :end_time
-    rename_column :inspections, :inspection_date, :inspection_time
   end
 
   def down
     rename_column :events, :start_time, :start_date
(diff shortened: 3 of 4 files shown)


kgf:
I may have uncommited changes ? I'm confused by this too.

houhoulis:
puzzled

Although I don't know what to make of the changes, as far as I can tell, they bring the database in line with the existing migrations in the repo.  The new "messages" columns match what's in the only migration that touches "messages" (which is from 2013).  And there are no migrations that mention tables called "notifications", "recipients", or "responses".

I don't see those tables referenced in the code either.  I'd say this PR is a viable merge candidate except for the mystery.

houhoulis:
@kgf

This PR reverts some changes that were added in this May 2015 commit: `56c11496b`.

That commit included one migration change, and several changes to `schema.rb`.

The migration change was an edit of a 2014 migration, making a change to one column in Channel.

The `schema.rb` changes contained much more than the changed column in Channel.  The rest of the changes seem extraneous.  The extraneous changes match what this PR is reverting.

### How this happened
1. Migrations make database changes as written, and then run `db:schema:dump` (I think).
2. `db:schema:dump`, whether run manually or as part of `db:migrate`, dumps the current schema in the database to the `db/schema.rb` file.

Regardless of how extraneous changes get into the database, once they're there, when the developer runs later migrations or runs `db:schema:dump`, the extraneous changes get inserted into `schema.rb`.  (Here's [the branch](https://github.com/houhoulis/lims/tree/migrations_versus_schema_changes) where I tried this out.)

My theory for the extraneous artifacts in schema.rb currently in `development`:

They were created via migrations that were then deleted from the repo without being rolled back first.  Either the migrations were run from a branch that was not merged into development, or development has been rebased (or equivalent changes happened to lose git history) such that there's no trace of the migration files.

Alternatively, the database could have been edited directly, via `psql`.  It would have the same eventual effect on `db/schema.rb`.

### Where we're at

From current `development`, if you run migrations on a fresh database...

... well, they blow up.  That's because of other, unrelated irregularities in our git history of migrations.

I just added fixes for those broken, unrelated migrations to this PR.  With those fixes added to `development`, running migrations on a fresh database produce the schema.rb changes in this PR.

### Why fix `schema.rb`

When a new developer joins the project, to set up their database, they will either:
1. run `rake db:schema:load`.
   1. The extraneous changes will be incorporated into their database.
   2. They'll look up to date with development.  Great, they're ready to go.
   3. _The changes have propagated.  They're a virus!_
2. run migrations.
   1. They'll have the schema.rb from this PR, not from development.  They'll look out of date with development & not ready to go.
   2. Every time they check out the development they'll get the extraneous-change `schema.rb`.  But any time they run a new migration they'll end up with these same changes as in this PR.

I think the virus needs to be expunged.  (Step 2: I think we should stop editing past migrations.)

houhoulis:
To set up their database, a new developer might also:

\3. run `rake db:setup`

I believe `db:setup` calls `db:schema:load` under the hood, so it's the same as 1. above.

PR #82, opened 2016-02-08
78 add resourcetype to items

kgf:
An effort was made to give more atomic commits.

### DIFF
--- a/app/views/items/show.html.erb
+++ b/app/views/items/show.html.erb
@@ -6,7 +6,7 @@
 <% end %>
 <p>
   <b>Location:</b>
-  <%= @item.location_id %>
+  <%= @item.location %>
   <h3>
     <%= @item.name + " ( " + @item.brand + " " + @item.model + " )" %></br>
     <span class="badge <%= 'badge-success' if @item.status == 'Available' %>">
--- a/spec/factories/resourcetypes.rb
+++ b/spec/factories/resourcetypes.rb
@@ -1,0 +1,10 @@
+FactoryGirl.define do
+  factory :resourcetype do
+    name "MyString"
+femakind "MyString"
+femacode "MyString"
+status "MyString"
+description "MyText"
+  end
+
+end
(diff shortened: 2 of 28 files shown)


houhoulis:
Besides fixing the merge conflict, could you redo this PR to use the more conventional compound word form?  Ruby typically would use "ResourceType" & "resource_type", rather than "Resourcetype" & "resourcetype".

I think "fema_kind" and "fema_code" would also be more conventional.

app/views/items/show.html.erb
```diff
@@ -6,7 +6,7 @@
 <% end %>
 <p>
   <b>Location:</b>
-  <%= @item.location_id %>
+  <%= @item.location %>
```
houhoulis, line 9: Would `@item.location.name` (or `.description`?) be more helpful here?

spec/factories/resourcetypes.rb
```diff
@@ -0,0 +1,10 @@
+FactoryGirl.define do
+  factory :resourcetype do
+    name "MyString"
+femakind "MyString"
+femacode "MyString"
+status "MyString"
+description "MyText"
```
houhoulis, line 7: Could you fix this indenting?  (Although, if you go with my advice about conventional compound word usage like "ResourceType", "fema_kind", etc, then this should resolve itself. :sunglasses: )

houhoulis:
Thanks for your effort making this a cohesive PR!  :+1: 

If you're willing to redo your naming (Resource Type, fema_kind, fema_code), then your attention to atomic commits is going to pay off right now!  This is how I would do it:
1. roll back the migrations in this PR
2. check out development
3. create a new branch
4. do the same scaffolding & migrations that you did here, except with the new naming
5. cherry-pick the last commit from this PR:
   1. `git cherry-pick 1ba7cc0`
   2. deal with a couple small merge conflicts & `git add` those files once conflicts fixed
   3. `git cherry-pick --continue`
6. Make new PR (or else reset your `78_add_resourcetype_to_items` branch to the new one and then force push it)

kgf:
Superseded by #85

Issue #78, opened 2016-01-31
Items need to track ResourceType


kgf:
FEMA ICS uses a resource type code. We need to capture that code for each item. The choices for the item should come from instances of a different model. This should include a _status_ field, so that ResourceTypes can be turned off if they do not apply.

PR #85, opened 2016-02-14
78 resource type part two

kgf:
A second, cleaner attempt at adding ResourceType for Items.
Resource Type refers to the official FEMA type system, used in ICS

This addresses Issue #78

### DIFF
--- a/app/helpers/resource_types_helper.rb
+++ b/app/helpers/resource_types_helper.rb
@@ -1,0 +1,2 @@
+module ResourceTypesHelper
+end
--- a/db/migrate/20160214032514_add_resource_type_id_to_item.rb
+++ b/db/migrate/20160214032514_add_resource_type_id_to_item.rb
@@ -1,0 +1,14 @@
+class AddResourceTypeIdToItem < ActiveRecord::Migration
+  def up
+    if column_exists? :items, :resourcetype_id
+      remove_column :items, :resourcetype_id
+    end
+    unless column_exists? :items, :resource_type_id
+      add_column :items, :resource_type_id, :integer
+    end
+  end
+
+  def down
+
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20160115231033) do
+ActiveRecord::Schema.define(:version => 20160214021906) do
 
   create_table "activities", :force => true do |t|
     t.string   "content"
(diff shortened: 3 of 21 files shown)


db/migrate/20160214032514_add_resource_type_id_to_item.rb
```diff
@@ -0,0 +1,14 @@
+class AddResourceTypeIdToItem < ActiveRecord::Migration
+  def up
+    if column_exists? :items, :resourcetype_id
+      remove_column :items, :resourcetype_id
```
masukomi, line 4: going with the assumption that the choice to not migrate the data is because there is no real data yet.

kgf, line 4: correct

db/schema.rb
```diff
@@ -187,14 +199,16 @@
...
-    t.datetime "created_at", :null => false
-    t.datetime "updated_at", :null => false
+    t.string   "channel"
+    t.datetime "processed_at"
+    t.string   "processed_by"
+    t.string   "slug"
+    t.datetime "created_at",   :null => false
+    t.datetime "updated_at",   :null => false
+    t.text     "body"
+    t.string   "subject"
```
houhoulis, line 211: Unfortunately these changes to "messages" are regressions, as are the reintroduction of the "notifications", "recipients", and "responses" below.  Also, below, two "resource type" tables are introduced.  We want the "resource_types" one.

Fortunately, we already investigated & solved this problem in #81, so we can fix this again.  We Have The Technology.  :runner:

spec/factories/resource_types.rb
```diff
@@ -0,0 +1,10 @@
+FactoryGirl.define do
+  factory :resource_type do
+    name "MyString"
+status "MyString"
+description "MyText"
+fema_code "MyString"
+fema_kind "MyString"
```
houhoulis, line 7: These weird outdents were also in #78.  I don't know why.

db/schema.rb
```diff
@@ -83,6 +83,16 @@
...
+  create_table "departments", :force => true do |t|
+    t.string   "name"
+    t.string   "shortname"
+    t.string   "status"
+    t.integer  "contact_id"
+    t.text     "description"
+    t.datetime "created_at",  :null => false
+    t.datetime "updated_at",  :null => false
+  end
+
```
houhoulis, line 95: I don't know where "departments" came from, but it should not be in this PR.

The miscellaneous introductions of incorrect schema changes are most likely coming from running migrations in other branches, and not rolling them back when switching branches.  I explained this a bit more in https://github.com/kgf/lims/pull/81#issuecomment-181159214

Editing previous migrations can have a similar effect.

Since this kind of undesirable schema change has happened in two recent PRs -- with this PR being worse than #78 -- I hope you can appreciate the importance of managing these schema changes carefully!  We can discuss further if you have questions.  We'll clean this up in another PR.

app/helpers/resource_types_helper.rb
```diff
@@ -0,0 +1,2 @@
+module ResourceTypesHelper
+end
```
houhoulis, line 2: Perhaps we could delete this empty module?  (Not critical to undo this addition.  It's more of a philosophical preference, which may not be shared.)

Issue #86, opened 2016-02-15
Filter locations for items based on department


kgf:
When entering an item, after selecting a department, only that departments locations should be available.

kgf:
This is not yet finished, as the filtering does not yet happen.

pragyajalan:
Pull request for this issue is merged - so closing this issue

Issue #69, opened 2016-01-18
Items need to have department


kgf:
Items could belong to the highway department or the Department of Public Works. The item should know what department it belongs to.
This should be a required field. Department should be a new object.
Department menu can go under admin

houhoulis:
@kgf An interesting idea is to add the specs for what you are asking for.  See https://medium.com/@kentcdodds/first-timers-only-78281ea47455.

@DJG86g don't wait though.  Just an idea that was on my mind, I'm not obligating @kgf  :smile:

PR #87, opened 2016-02-15
69 add department to item

kgf:
Addresses #69 
Could use some additional polish around switching to protected attributes, and filtering choices. This has been filed as issue #86


app/models/location.rb
```diff
@@ -1,4 +1,6 @@
 class Location < ActiveRecord::Base
-  attr_accessible :category, :comments, :description, :lat, :lon, :name, :status
+  attr_accessible :category, :comments, :description, :department_id,
+                  :floor, :container, :lat, :lon, :name, :status
```
masukomi, line 3: the non-department stuff probably shouldn't be in this pr. will merge anyway but...

PR #90, opened 2016-02-18
Remove occasional requires of absent rails_helper.

houhoulis:
`rails_helper.rb` is reportedly the preferred RSpec way going forward, but these 5 includes were a bit hasty.


houhoulis:
This does not fix the specs, but it is a start.

houhoulis:
I am rolling this fix into PR #91 , partly to see what Travis CI does with it.

PR #91, opened 2016-02-21
Fix specs and 'item belongs_to owner' association.

houhoulis:
This PR fixes several recent additions that were incomplete.  It restores all specs to green.  It also fixes some deprecated RSpec syntax, cleans up our database_cleaner config, and adds :green_heart: Travis CI! :heart: :green_heart:


houhoulis:
Do not merge this PR yet.  It's intentionally incomplete so that we can see how Travis handles our various failing (or bombing-out) tests.

kgf:
Interesting, 11 runs already.

houhoulis:
The only test failure is in ruby 2.3.  Props to Travis because I ran the specs locally in ruby 2.3 and duplicated the failure.

It's a strange failure.  I'm not sure whether the cause is the spec or RSpec + ruby.

Either way, I have a fix for the spec.  Will push up tomorrow.

houhoulis:
This bug has nothing to do with RSpec.  Rather, the "`conditions`" on the `Person.certs` scope is respected in ruby 2.1, and ignored in 2.3:

``` ruby
2.1.8:001>p = Person.create! firstname:"f",lastname:"l",status:"Active"
2.1.8:002>p.certs
  Cert Load (1.0ms)  SELECT "certs".* FROM "certs" WHERE "certs"."person_id" = 3 AND "certs"."status" = 'Active'
```

``` ruby
2.3.0:001>p = Person.create! firstname:"f",lastname:"l",status:"Active"
2.3.0:002>p.certs
  Cert Load (0.5ms)  SELECT "certs".* FROM "certs" WHERE "certs"."person_id" = 4
```

Fascinating.

houhoulis:
So.

[This](https://github.com/kgf/lims/blob/04e7201da2c5fd79d3bcf1531c34b373f415ba74/app%2Fmodels%2Fperson.rb#L5) is our association in Person:

`has_many :certs, :conditions => {:status =>'Active' }`

In Rails 3.2, you can define conditions on an association with a String + interpolation, such as `conditions: ["status = ?", 'Active']`.  You can also do it with a Hash as we've done here: `conditions: { status: 'Active' }`.

The benefit of using a Hash is that `person.certs.build` or `person.certs.create` will build or create a Cert object with status "Active".  The String interpolation version, by contrast, does not pre-populate `status`.

Since we've used a Hash here, I am assuming we want that added behavior.

The problem is, Ruby 2.3 changes `Hash` to accept a `to_proc` method.  Which is cool.  But it causes a regression in Rails v3.2: https://github.com/rails/rails/pull/23082  The `conditions` hash is not applied to the association scope, it seems.

Fortunately, we can define a lambda for `conditions` as a variant of the Hash version.  (An example is in the rails PR I just linked.)  The lambda seems to preserve the behavior of the hash version, and it seems to work correctly in Ruby 2.1 and 2.3.

I did look into implications for Rails 4+, not wanting to complicate our upgrade path.  The docs suggest that using the explicit `:conditions` option isn't supported at all: http://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many .  I'm not sure what syntax will be allowed as deprecated, but changing this association to use a lambda isn't clearly more complicated to migrate.

I will change the association to use a lambda.

houhoulis:
Food for thought: the example in [a comment](https://github.com/rails/rails/pull/23082#issuecomment-176629716) on the rails PR would, in our domain, apply the "`conditions`" lambda on `certs` to the `has_many skills` association, rather than to the `has_many certs` association.

We could define additional scopes, and/or move the "`conditions`" clause, if we wanted to be able to see all a user's certs, courses, or skills rather than just the "Active" ones.

We might want to consider whether to move restrictions, like this one on `cert.status`, into named scopes/associations rather than on the direct association.  The fact that `Person.first.certs` doesn't return all Certs where `person_id == Person.first.id` always has the capacity to surprise the unwary developer.

Issue #80, opened 2016-02-01
Items will need an ItemType


kgf:
This will be used to allow decisions on inspections and grouping. 
For example if there is a truck of type `Ford F-250`, you would have inspection questions about checking the transmission fluid, and you could group it with a snowplow mount. An item with the Type of `Towable Lighting Unit` would include inspection questions about the condition of the mast cables, and would not accept any groups. Model should include status, so a type can still exist, but not be available for new items.

PR #95, opened 2016-03-12
Adding item_types as a model

kgf:
Adding ItemType to Items, so we can group and categorize them

### DIFF
--- a/spec/factories/item_types.rb
+++ b/spec/factories/item_types.rb
@@ -1,0 +1,10 @@
+FactoryGirl.define do
+  factory :item_type do
+    name "MyString"
+    status "MyString"
+    is_groupable "MyString"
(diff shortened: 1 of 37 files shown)


kgf:
Waiting to see if tests pass, then I will add to this PR

spec/factories/item_types.rb
```diff
@@ -0,0 +1,10 @@
+FactoryGirl.define do
+  factory :item_type do
+    name "MyString"
+status "MyString"
+is_groupable "MyString"
+is_a_group "MyString"
+parent_id 1
```
houhoulis: why are these fields aligned all the way to the left, do you know?

kgf:
This addresses issue #80

PR #45, opened 2015-05-06
43 ruby rmagick versions

DJG86g:
changes made to lock rmagick version as well as downgrading to ruby version from 2.2.2 to 2.1.2 .

PR #102, opened 2016-04-10
Upgrade rmagick gem to 2.15.4.

houhoulis:
This version has been easier to install for some people.  It also
fixes a segfault bug which I've encountered in a different project.


houhoulis:
If accepted, this PR would supercede #45 .

PR #103, opened 2016-05-02
89 Much clean-up of Item show, including UniqueIds

kgf:
Tests pass, and code is usable. Not where I would like it to be, but usable.

Also, hid many fields if they had no value. Stripped out empty UniqueIds, preventing them from being saved.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,2 +1,10 @@
 module ApplicationHelper
+  def link_to_add_fields(name, f, association)
+    new_object = f.object.send(association).klass.new
+    id = new_object.object_id
+    fields = f.fields_for(association, new_object, child_index: id) do |builder|
+      render(association.to_s.singularize + "_fields", f: builder)
+    end
+    link_to(name, '#', class: "add_fields", data: {id: id, fields: fields.gsub("\n", "")})
+  end
 end
--- a/app/views/items/_form.html.erb
+++ b/app/views/items/_form.html.erb
@@ -4,6 +4,7 @@
   <div class="form-inputs">
     <%= f.association :department %>
     <%= f.association :location %>
+    <%= f.association :owner %>
     <%= f.input :status, :as => :select, :collection => Item::STATUS_CHOICES %>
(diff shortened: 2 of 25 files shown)


app/views/items/_form.html.erb
```diff
@@ -12,9 +13,9 @@
     <%= f.input :model %>
     <%= f.input :stock_number %>
     <%= f.input :description %>
-    <%= f.input :serial1, :label => 'Serial Number 1' %>
-    <%= f.input :serial2, :label => 'Serial Number 2' %>
-    <%= f.input :serial3, :label => 'Serial Number 3' %>
+    <%#= f.input :serial1, :label => 'Serial Number 1' %>
+    <%#= f.input :serial2, :label => 'Serial Number 2' %>
+    <%#= f.input :serial3, :label => 'Serial Number 3' %>
```
houhoulis: Could you remove these lines?

app/views/items/show.html.erb
```diff
@@ -5,58 +5,53 @@
...
 <% end %>
 <p>
-  <b>Location:</b>
-  <%= @item.location_id %>
   <h3>
-    <%= @item.name + " ( " + @item.brand + " " + @item.model + " )" %></br>
     <span class="badge <%= 'badge-success' if @item.status == 'Available' %>">
       <%= @item.status %>
-    </span>
+    </span>&nbsp
```
houhoulis: I think you are missing a `;` at the end.

houhoulis:
You've added an empty UniqueIdHelper spec... and an empty UniqueIdHelper class.  Could we just get rid of them?

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,10 @@
 module ApplicationHelper
-end
\ No newline at end of file
+  def link_to_add_fields(name, f, association)
+    new_object = f.object.send(association).klass.new
+    id = new_object.object_id
+    fields = f.fields_for(association, new_object, child_index: id) do |builder|
+      render(association.to_s.singularize + "_fields", f: builder)
```
houhoulis, line 6: Fancy!

Issue #13, opened 2013-11-06
Allow upload of images to a person for authentication


kgf:
This is basically a passport style photo. By Authentication, I mean that someone looking at our system knows who they are looking at in front of them.

kgf:
Added Carrierwave in PR #120

PR #113, opened 2016-08-07
Added Carrierwave for a portrait for person

kgf:
This addresses issue #13

### DIFF
--- a/app/uploaders/portrait_uploader.rb
+++ b/app/uploaders/portrait_uploader.rb
@@ -1,0 +1,51 @@
+# encoding: utf-8
+
+class PortraitUploader < CarrierWave::Uploader::Base
+
+  # Include RMagick or MiniMagick support:
+  # include CarrierWave::RMagick
+  # include CarrierWave::MiniMagick
+
+  # Choose what kind of storage to use for this uploader:
+  storage :file
+  # storage :fog
+
+  # Override the directory where uploaded files will be stored.
+  # This is a sensible default for uploaders that are meant to be mounted:
+  def store_dir
+    "uploads/#{model.class.to_s.underscore}/#{mounted_as}/#{model.id}"
+  end
+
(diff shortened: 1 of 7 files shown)


app/uploaders/portrait_uploader.rb
```diff
@@ -0,0 +1,51 @@
+# encoding: utf-8
+
+class PortraitUploader < CarrierWave::Uploader::Base
+
+  # Include RMagick or MiniMagick support:
+  # include CarrierWave::RMagick
+  # include CarrierWave::MiniMagick
+
+  # Choose what kind of storage to use for this uploader:
+  storage :file
```
JosephWinfield, line 10: Are we going to set this up with an s3 bucket? I think if there are enough users or just people in the app it might lag.

db/migrate/20160807063212_add_portrait_to_person.rb
```diff
@@ -0,0 +1,5 @@
+class AddPortraitToPerson < ActiveRecord::Migration
+  def change
+    add_column :people, :portrait, :string
+  end
+end
```
JosephWinfield, line 5: no down method?

app/uploaders/portrait_uploader.rb
kgf: I don't want to make the assumption that people are running this only on aws. Adding a note somewhere in the docs would be good though.

db/migrate/20160807063212_add_portrait_to_person.rb
kgf: Fair enough. I'll add one.

kgf: Whoops, this is a change method, not an up. Change means rails knows how to do the up and the down.

kgf:
Between repo, conflicts not resolved

PR #115, opened 2016-08-09
Rails40 - This will hopefully get things working for Rails 4.0

david135:
This will hopefully get things working for Rails 4.0.  Further updates, toward either 4.2.7 or 5.0, depending on your preferences, will be in the works. The commits are intended to be somewhat free-standing, and if need be you may be able to cherry-pick from them to a certain extent.

A few of the commit messages contain a special request for review, but any is appreciated. I'll be happy to see any comments or suggestions for improvement.

A few points:
  1) the newer version of devise called for a secret key.  Since that involves security, please do take a look.
  2) The only remaining deprecation (of those that emerge from the RSpec tests - I suspect that there are more, such as uses of 'all') is about setting 'config.secret_key_base'.  It seems that this will need to wait until the production database is ready to use the new cookie system ([A Guide for Upgrading Ruby on Rails](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html)):

> Please note that you should wait to set secret_key_base until you have 100% of
> your userbase on Rails 4.x and are reasonably sure you will not need to rollback to Rails 3.x.

  3) I used [arel_converter](https://github.com/pallan/arel_converter) for some of the models.  It's worth checking that over, as I'm not expert in Rails 3 syntax.

  4) `bootstrap-responsive` is no longer separate.  I haven't tested my changes extensively, and for actual use it should definitely be tested and improved.

  5) `bootstrap-sass` is in need of updating, but I wanted to see how this is looking before diving in.

All the tests pass, and basic usage seems to work in the web browser.  I think we'd benefit from the db seeds that are being discussed, as well as descriptions of standard user stories for developers to keep in mind.

Thanks!


kgf:
Sorry for the delay in looking at this, I am currently on vacation.

kgf:
The top menus do not seem to work. Would you be able to get them working, or a reasonable substitution ?
Thanks for your help !

screen shot 2016-08-16 at 2 16 37 am

david135:
Thanks for taking a look (and enjoy your vacation)! Could you please tell me which browser/OS you're using (and/or any other advice you might have) to reproduce the problem?  I have menus working for me in Firefox 46.0.1 and Chrome 52.0.2743.116 (image attached), both OS X.

BTW, I see that the table has an ordering issue.  Would you prefer "Name Model Location", as in the header or "Location Name Model" as in the item display (views/items/index.html.erb)?

lims40-menu-chrome

PR #116, opened 2016-08-19
Rails42 - This should be working for Rails 4.2.7

david135:
This should bring us to Rails 4.2.7. It passes the tests and seems to work in the browser, though as before it will benefit from a well-defined user story and a seeded db.

Some notable changes:
1. Bootstrap is now working at version 2.3.2.  Two Bootstrap changes of note: 1) the default color isn't black anymore, so `navbar-inverse` is necessary in `_top_menu.html.erb`. 2) Bootstrap no longer preserves the empty column if the sidebar is empty. If this is a problem, we'll need a workaround for many of the views.
2. Made many small changes to bring the Rails boilerplate in line with the current rails releases (based on railsdiff.org).
3. Also made several other fixes and cleaned up/reduced scss files.
4.  Now using the `jquery-turbolinks` gem with the `turbolinks-compatibility.coffee` shim to get turbolinks working with jquery.
5. Added `Rails.application.config.assets.precompile += %w( glyphicons-halflings.png glyphicons-halflings-white.png` to assets.rb
6. Made Gemfile more similar to default Gemfile (in terms of order) in preparation for continuing work.

TODO (partial):
1. `DEPRECATION WARNING: You didn't set 'secret_key_base'. Read the upgrade documentation to learn more about this new config option.`
2. `config/secrets.yml` to replace `config/initializers/secret_token.rb` (cf. above)
3. `<h1></h1>`in the sidebar doesn't look very good. (e.g. app/views/users/edit.html.erb)
4. Clean up, update, and organize Gemfile.
5. Update `bootstrap` to Bootstrap 3 or Bootstrap 4?
6. bring `config/routes.rb` up to date
7. `config/database.yml`
8. `config/initializers/cookies_serializer.rb`


app/assets/stylesheets/departments.css.scss
```diff
@@ -1,3 +0,0 @@
-// Place all the styles related to the Departments controller here.
```
kgf, line 1: Thanks for the clean-up of the empty files

config/environments/development.rb
```diff
@@ -1,33 +1,44 @@
...
+  # Do not eager load code on boot.
+  config.eager_load = false
+
+  # Show full error reports and disable caching.
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false
 
-  # Don't care if the mailer can't send
+  # Don't care if the mailer can't send.
   config.action_mailer.raise_delivery_errors = true
```
kgf, line 17: Is this duplicated on line 43 ?

config/environments/production.rb
```diff
@@ -1,69 +1,85 @@
...
-  # Prepend all log lines with the following tags
+  # Prepend all log lines with the following tags.
   # config.log_tags = [ :subdomain, :uuid ]
 
-  # Use a different logger for distributed setups
+  # Use a different logger for distributed setups.
   # config.logger = ActiveSupport::TaggedLogging.new(SyslogLogger.new)
 
-  # Use a different cache store in production
+  # Use a different cache store in production.
```
kgf, line 63: Seems like a lot of changes in this area are just adding a period, or changing the spelling of Apache. Was this something that came with Rails ?

config/initializers/secret_token.rb
```diff
@@ -1,7 +1,8 @@
 # Be sure to restart your server when you modify this file.
 
-# Your secret key for verifying the integrity of signed cookies.
+# Your secret key is used for verifying the integrity of signed cookies.
 # If you change this key, all old signed cookies will become invalid!
+
 # Make sure the secret is at least 30 characters and all random,
 # no regular words or you'll be exposed to dictionary attacks.
 Lims3::Application.config.secret_token = 'a4894c186c085cabe496af731838f541fb664763ed0900e9cee471790f0dd734965817bf10d36401ff7d72e86d62146a96844dc356b809329be2f71aca48fdec'
```
kgf: Does this reference to LIMS need to be changed to `Rails.Application` ?

config/initializers/session_store.rb
```diff
@@ -1,8 +1,3 @@
 # Be sure to restart your server when you modify this file.
 
-Lims3::Application.config.session_store :cookie_store, key: '_lims3_session'
-
-# Use the database for sessions instead of the cookie-based default,
-# which shouldn't be used to store highly confidential information
-# (create the session table with "rails generate session_migration")
-# Lims3::Application.config.session_store :active_record_store
+Rails.application.config.session_store :cookie_store, key: '_lims3_session'
```
kgf: Should we change the key, as `lims3` would be confusing.

config/environments/development.rb
david135: Sure is.  Turns out it's been duplicated since c55e3a8 in January 2013.  The earlier one is where it's found in a fresh rails app, while the latter one is with the other email configuration added for this app.

config/environments/production.rb
david135: Indeed.  Because the plan was to upgrade 3.2->4.0->4.1->4.2(->5.0), I wanted to reduce differences from a fresh app to help avoid missing necessary changes. Most/all of the changes in comments are of that type, based on railsdiff.org's listings.  If you prefer to not have these, I think I could undo them, but this does make it easier to see what's different about this app compared to another. Let me know.

kgf: If this bings it closer in line to a standard Rails app, better to have that pain now.

config/initializers/session_store.rb
kgf: Please change the key to `_rdy_resp_session`

config/environments/development.rb
kgf: Thanks. Please remove the latter one. DRY.

kgf:
Other than those small changes, I think we're good to merge.

Does this PR also address #117  ?

config/initializers/secret_token.rb
david135: Thanks, likely it should, but I had left it this way primarily because we should get rid of this file entirely and use secrets.yml instead. I have a brief note on this in this PR and the rails40 PR, but didn't want to change security procedure without your OK.

config/initializers/session_store.rb
david135: Shall we also rename the app to ReadyResponder in `config/application.rb`?

david135:
Yes, this should resolve #117, choosing to fix the header to match the data.

PR #120, opened 2016-08-25
13 add portrait to person

kgf:
This only takes a single image, not yet a relation allowing many photos


kgf:
To the reviewer, please follow the travis link and observe that the tests did pass for Ruby 2.3. Apparently we now have a dependency on that, which is OK.

app/views/people/_form.html.erb
```diff
@@ -48,6 +48,9 @@
     <%= f.input :start_date, :as => :string, :input_html => { :class => 'datepicker', :size => 11 }  %>
     <%= f.input :division1, :as => :select, :label => 'Division', :collection => Person::DIVISION1 %>
     <%= f.input :division2, :as => :select, :label => 'Squad', :collection => Person::DIVISION2 %>
+    <h4> Personnel Image </h4>
+    <%= @person.portrait %>
```
david135, line 52: personnel image filename
As it is now, when a person has a portrait, it shows the portrait's filename, but if someone already has a portrait, it might be better to show the picture itself), with the option to change the picture, and possibly to delete it.

app/views/people/_form.html.erb
```diff
@@ -48,6 +48,9 @@
...
+    <%= f.file_field :portrait %>
     <%= f.input :comments, :input_html => { :size => 40}  %>
```
david135, line 54: personnel image comments and browse button
With the addition of the `<h4>` tag for the Personnel Image, it now looks like the comments field is designed to go with the Personnel Image. I think it needs at least a header of the same level as Personnel Image, but there could be some more comprehensive styling consistency on this page.

app/views/people/_form.html.erb
```diff
@@ -48,6 +48,9 @@
     <%= f.input :start_date, :as => :string, :input_html => { :class => 'datepicker', :size => 11 }  %>
     <%= f.input :division1, :as => :select, :label => 'Division', :collection => Person::DIVISION1 %>
     <%= f.input :division2, :as => :select, :label => 'Squad', :collection => Person::DIVISION2 %>
+    <h4> Personnel Image </h4>
+    <%= @person.portrait %>
+    <%= f.file_field :portrait %>
```
david135, line 53: For some reason I have a broken "Browse..."  button displayed.  It is missing the bottom line. See pictures above.

app/views/people/show.html.erb
```diff
@@ -1,7 +1,8 @@
 <% content_for :sidebar do %>
   <%= link_to 'Return to Listing', session[:return_to], :class => 'btn btn-mini' %></br></br>
   <%= link_to 'Edit Person', edit_person_path(@person), :class => 'btn btn-mini' if can? :edit, @person  %></br></br>
-  <%= link_to 'New Certification', new_person_cert_path(@person), :class => 'btn btn-mini' if can? :create, Cert  %>
+  <%= link_to 'New Certification', new_person_cert_path(@person), :class => 'btn btn-mini' if can? :create, Cert  %></br></br>
+  <%= image_tag @person.portrait_url if @person.portrait.file.present? %>
```
david135, line 5: Should we have an option to follow a link to a full-size image here? In a large browser window, I'm showing a thumbnail-sized picture in the sidebar, while in a smaller window or on an iPhone I see a full picture.

Large Window:
person display - browser large

Smaller Window:
person display - browser small

iPhone:

person display - iphone

david135:
Do we want tests for a new feature like this?  It works in my testing-by-hand.

david135:
Do we want to add `public/uploads/` to `.gitignore`? Is the plan to store uploaded images locally?

david135:
Should `db/schema.rb` be committed in this PR?

app/uploaders/portrait_uploader.rb
```diff
@@ -0,0 +1,51 @@
...
+  # end
+
+  # Create different versions of your uploaded files:
+  # version :thumb do
+  #   process :resize_to_fit => [50, 50]
+  # end
+
+  # Add a white list of extensions which are allowed to be uploaded.
+  # For images you might use something like this:
+  # def extension_white_list
```
david135: This white list may be a good idea.

david135:
See also issue #125.

app/views/people/_form.html.erb
kgf: I would expect the image to show in the left hand column

app/views/people/_form.html.erb
kgf: Agreed, the styling is bad, bad, bad.

app/views/people/show.html.erb
kgf: The thumbnail should be a link to the largest size available

app/uploaders/portrait_uploader.rb
kgf: I will add the whitelist to the PR. I will probably also move the images out of public to make security better.

kgf:
I did not implement some of the scaling features available in Carrierwave. There is also no way to delete an image through the UI.

PR #122, opened 2016-08-25
Removed Ruby 2.1 from the Travis test builds

kgf:


david135:
This does solve the problem with PR #120, and we don't need to support older Rubies if there's no benefit, but for the future, do we also want to consider using `bundle` without the development group (`--without=development`) in travis? This may require running through the Gemfile to make sure all gems are still categorized properly, but that may be a good idea anyway.  Am I right that several gems that are in :test and :development could be in one or the other.  The dependencies run: guard->listen->ruby_dep, which is what failed in this case (cf. [the first paragraph here](https://github.com/guard/listen/wiki/Support-for-older-versions-of-Ruby#myth-3-that-failing-citravis-builds-means-semver-rules-were-violated)).

david135:
Am I right that we should either update `.ruby-version` or remove it? It seems that this would change in sync with `.travis.yml` (`git diff 249b009~1 249b009`).  We could also be moving to ruby 2.3.1 here, or else have both 2.3.0 and 2.3.1 on the list. Otherwise, This seems ready to merge.

kgf:
I agree, I will amend this pr to include Ruby version.
I would like the bundle changes to be in a different pr, but I like the idea.

Issue #124, opened 2016-08-25
Dropdown menus don't work in iOS


david135:
I've found a temporary solution that works for iOS, which works if placed in app/assets/bootstrap232_dropdown_menu_fix.js:

```
// This fix is for Bootstrap 2.3.2 on iOS to get the dropdown menus working.
// http://stackoverflow.com/questions/17178606/bootstrap-v2-dropdown-navigation-not-working-on-mobile-browsers
  jQuery(document).ready(function($) {
    $("li.dropdown a").click(function(e){
      $(this).next('ul.dropdown-menu').css("display", "block");
      e.stopPropagation();
    });
  });
```

I have no idea whether this fouls anything up for Android.

It seems, though, that the better solution will be to upgrade Bootstrap. Would we want something like this for now?

kgf:
I would prefer to fix it dead, rather than a hack.

david135:
I'm working on seeing if Bootstrap 3 fixes this.

david135:
This seems to be resolved.  Please report if problems arise.

Issue #123, opened 2016-08-25
Cancel button for people, which uses session[:return_to], is broken


david135:
in `app/views/people/edit.rb` and elsewhere:

```
<%= link_to 'Cancel', session[:return_to], :class => 'btn' %></br>
```

But `session[:return_to]` isn't set properly when going from http://localhost:3000/people/everybody and clicking the edit button.

See: http://stackoverflow.com/questions/2139996/how-to-redirect-to-previous-page-in-ruby-on-rails, where they write:

> In your edit action, store the requesting url in the session hash, which is available across multiple requests:
> `session[:return_to] ||= request.referer`
> Then redirect to it in your update action, after a successful save:
> `redirect_to session.delete(:return_to)`

It seems that it is never deleted, and so it is never re-set.  See people_controller.rb:

```
  def set_return_path
    session[:return_to] ||= request.referer
  end
```

We set it as recommended, but we never clear the value, so whenever it is set once, it sticks and is never updated. (Until we reset the cookie signing, this had been going to a random, unrelated page, but now it seems to be set to the root page, because that is the first and only time the value is null, and therefore set when we go through `PeopleController` the first time.)

This seems to go back to commit 7572cc7f89c84fba4045a2cf80ef4555a92fd0a7.

kgf:
@david135 In your research, what would be the solution to this ?

david135:
It seems that the idiom is to delete `session[:return_to]` every time we go to it, then leaving it ready to be re-set.  Having the views (edit, new, show) link to `session[:return_to]` makes it a little complicated. I would use a helping function.  If we compare `view/items/new.html.erb` and `view/people/new.html.erb`, we see the difference in behavior.  Do we want this functionality for all models ('cancel' returns us to wherever we were when we selected 'New (Person|Item|Event)')? If so, abstracting that behavior out of the people_controller may be the way to go.

david135:
The cancel button is now working, since the above-mentioned PRs, but this is still a problem in the controller, setting and using `session[:return_to]`, which impacts redirects.

kgf:
@david135 should we close this issue, and open a new one that is more specific ?

david135:
Resolved in PR #197.

Issue #119, opened 2016-08-24
Menus do not show properly when adding a new event.


kgf:
I believe this was already fixed elsewhere by David, as part of the Rails 4.2 upgrade
<img width="309" alt="screen shot 2016-08-24 at 7 19 17 pm" src="https://cloud.githubusercontent.com/assets/2431417/17951395/dd0d5278-6a2f-11e6-9f1c-0c4007be024a.png">

kgf:
This was addressed by #132

Issue #118, opened 2016-08-19
Inventory column titles do not scroll


kgf:
Please have inventory column titles freeze, so that the designations are readable when scrolling down the list.

kgf:
This will probably require this change to the jQuery Datatables
https://datatables.net/extensions/fixedheader/

kgf:
While the column headers do not freeze, there is now pagination, which should help.

david135:
They freeze for me.  If they don't for you, there is something to investigate.  This should be a feature of all the datatables, set in `datatables-defaults.js`:
`scrollY: '50vh'`
![scrolling table](https://cloud.githubusercontent.com/assets/8586431/18771423/aee2b0b6-810c-11e6-98f2-fc382bacfa79.gif)

PR #132, opened 2016-09-09
Bootstrap3

david135:
This PR is meant to bring us to Bootstrap 3.  It's been tested with OS X Firefox and iOS Safari.

Of note:
1. DataTables is now maintained manually in `vendor/assets/[stylesheets|javascripts]` instead through the `jquery-datatables-rails` gem.
2. We'll want to consider how to maintain/deploy this.  At a minimum, we'll want to use the minified versions.
3. DataTables now uses the scrollY option (keeping the header visible) and its Responsive feature, which will hide columns when the table is too wide for the screen. See configuration in:
   `datatables_defaults.js` as well as `items.js.coffee` and `people.js.coffee` for examples.  It defaults to hiding the rightmost columns, so when that is not desired, it is configured as in `items/people.js.coffee`.  Using [`columnDefs`](https://datatables.net/reference/option/columnDefs) instead of `columns`.
4. Most DataTables configuration is now set in a single place, `datatables_defaults.js`, though individual initialization is still needed for tables with any special column configuration.
5. Turbolinks is disabled, as it will take a good amount of work to make DataTables behave properly with it (it has never (since it was enabled a few PRs ago) worked properly with the back key, for example).
6. Use of `session[:return_to]` in views has been replaced with `:back`.
7. I removed `scaffolds.scss`, which confounded Boostrap links.  Let's watch out for anything that we rely on.
8. The `simple_form` initialization needs to be updated, which should improve its styling. They have a bootstrap variation if that suits us.
9. In `application.js`, the `require jquery-ui*` lines should be checked over.  Some may need to be added back or updated.
10. I made some educated guesses in `item_label` in `items_helper.rb`.  It'd be good for someone to check them over in terms of `success/warning/danger`.

This addresses #124, #123 (partially - in views, but not in the controller), #118, #119 (if it was still an issue), and hopefully some other concerns!


kgf:
This looks much better. One thing though, the search text box is gone. Config option ?

kgf:
There is now an issue when editing channels for a person. The sms tag does not line up. Perhaps the best solution is to drop this. How many cell phones DON'T have SMS anymore ?

screen shot 2016-09-09 at 7 03 15 pm

Gemfile.lock
```diff
@@ -58,13 +50,14 @@ GEM
       rack (>= 0.9.0)
     binding_of_caller (0.7.2)
       debug_inspector (>= 0.0.1)
-    bootstrap-sass (2.3.2.0)
-      sass (~> 3.2)
+    bootstrap-sass (3.3.7)
+      autoprefixer-rails (>= 5.2.1)
```
kgf, line 54: It looks like this is being required here by bootstrap-sass. Do we need to require it above ? Are we using it separately ?

app/assets/javascripts/application.js
```diff
@@ -11,17 +11,11 @@
...
 //
 //= require jquery
+//= require bootstrap-sprockets
 //= require jquery_ujs
-//= require jquery.turbolinks
-//= require turbolinks
-//= require turbolinks-compatibility
-//= require jquery-ui/core
+//= require datatables
+//  require turbolinks
```
kgf: It was mentioned in the PR comments that this was disabled. Do we still need it here ?

app/assets/javascripts/datatables_defaults.js
```diff
@@ -0,0 +1,21 @@
...
+/* https://datatables.net/examples/advanced_init/defaults.html */
+/* https://datatables.net/reference/option/ */
+
+$.extend( true, $.fn.dataTable.defaults, {
+  scrollY: '50vh',              // https://datatables.net/examples/basic_init/scroll_y_dynamic.html
+  responsive: true,             // https://datatables.net/extensions/responsive/priority
+  pagingType: "full_numbers",
+  retrieve: true,
+  pageLength: 100,
+  searching: false,
```
kgf, line 11: Ah, that's where the search box went 😉

app/assets/javascripts/items.js.coffee
```diff
@@ -4,3 +4,15 @@ jQuery ->
     regexp = new RegExp($(this).data('id'), 'g')
     $(this).before($(this).data('fields').replace(regexp, time))
     event.preventDefault()
+
+jQuery ->
+  $('#items').dataTable
+    columnDefs: [
+                  ## https://datatables.net/extensions/responsive/priority
```
kgf, line 11: I really appreciate this comment, makes review easier. And "future us" will probably use it too.

app/assets/javascripts/items.js.coffee
```diff
@@ -4,3 +4,15 @@ jQuery ->
...
+                  { responsivePriority: 1, targets: [0, 1, 2, 3, 8, 9 ] },
```
kgf, line 12: This is a good start, but we'll need to change the view itself. In a different PR, we should remove the `serial X` columns, as they don't really apply anymore. Also, I'd like to remove the `Available` status column, and just colorize the row itself.
https://datatables.net//forums/discussion/34882
I added a new issue...

app/helpers/application_helper.rb
```diff
@@ -8,3 +20,14 @@ def link_to_add_fields(name, f, association)
     link_to(name, '#', class: "add_fields", data: {id: id, fields: fields.gsub("\n", "")})
   end
 end
+
+# This is a general helper to provide a simple success or warning
+# label.  For anything more precise, a custom helper should be
+# used (as in items_helper.rb)
+def success_or_warning_label(str, success_str)
```
kgf, line 27: Interesting

app/views/courses/index.html.erb
```diff
@@ -1,4 +1,4 @@
-<table id="courses" class="datatable display table table-striped table-bordered" >
+<table id="courses" class="datatable table table-striped table-bordered" >
```
kgf, line 1: In other places you have changed this to `generic_datatable`. ( See `app/views/channels/index.html.erb:L6` ) Should this be the same ? What is the difference ?

app/views/events/_all_for_event.html.erb
```diff
@@ -5,9 +5,9 @@
       <th>Name</th>
       <th>Description</th>
       <th>Intention</th>
-      <th>Intented Start</th>
```
kgf, line 8: Oh github, what are you trying to tell me ?

app/views/items/show.html.erb
```diff
@@ -1,26 +1,27 @@
 <% content_for :sidebar do %>
-  <%= link_to 'Return to Listing', items_path, :class => 'btn' %></br></br>
-  <%= edit_item_button %></br></br>
-  <%= add_repair_button %></br></br>
-  <%= add_inspection_button %></br></br>
+  <%= link_to 'Return to Listing', items_path, class: 'btn btn-primary', style: 'width: 100%' %><br/><br/>
+  <%= edit_item_button %><br/><br/>
+  <%= add_repair_button %><br/><br/>
+  <%= add_inspection_button %><br/><br/>
```
kgf, line 5: I think you are more right than I was, but perhaps no slash would be best....
http://stackoverflow.com/questions/1946426/html-5-is-it-br-br-or-br

app/views/items/show.html.erb
```diff
@@ -1,26 +1,27 @@
...
 <% end %>
 
-<p><b>Location:</b> <%= @item.location %></p>
+<% if @item.location.present? %>
+  <p><b>Location:</b> <%= @item.location %></p>
+<% end %>
 
 <% if @item.source.present? %>
-  <p>Source:<b><%= @item.source %></b></p>
+  <p><b>Source:</b> <%= @item.source %></p>
```
kgf, line 24: It's no big deal, but I typically bold the content, not the labels. After a few times using it, people will be much more interested in the content, and will know where it lives.

app/views/layouts/_top_menu.html.erb
```diff
@@ -1,97 +1,100 @@
...
+        <span class="sr-only">Toggle navigation</span>
         <span class="icon-bar"></span>
         <span class="icon-bar"></span>
         <span class="icon-bar"></span>
-      </a>
-      <a class="brand" href="#">LIMS</a>
-      <div class="nav-collapse">
-        <ul class="nav">
+      </button>
+      <a class="navbar-brand" href="#">LIMS</a>
```
kgf, line 10: If you have time, could you change this to "Home", at least until we can make it configurable. With the name change, I don't want to leave it as LIMS

app/views/people/cert.html.erb
```diff
@@ -1,31 +0,0 @@
-<% content_for :sidebar do %>
```
kgf, line 1: I like deleting zombie code, thanks !

app/assets/javascripts/application.js
david135: Happy to remove it.  I had left it in hopes that we could figure out an approach to getting it working.

app/assets/javascripts/datatables_defaults.js
david135: Yes.  Would you like it back as a default, or do we want to add it for specific tables?

app/views/courses/index.html.erb
david135: IIRC, some tables were styled as generic_datatable to begin with, and were initialized by class (rather than by id) in `app.js.coffee`.  I moved that to `datatables_defaults.js` as part of this update.  Other datatables are initialized separately, as in this case (in `courses.js.coffee`), if they are customized instead of relying only on the defaults.  If I changed any, it (barring errors) was because they no longer had any customized behavior differing from the defaults.

app/views/events/_all_for_event.html.erb
david135: Just a little typo fix.

app/views/items/show.html.erb
david135: Happy to make a global switch if you would like as a separate commit.

app/views/items/show.html.erb
david135: I think I was attempting to make them consistent with line 47/8 below ("Expenses in last 6 months"), but that can go the other way too. Your call.

app/views/layouts/_top_menu.html.erb
david135: You bet.

Gemfile.lock
david135: It seems not.  I'll remove it.

app/views/layouts/_top_menu.html.erb
david135: It turns out we'll need to update 30 tests first!

david135:
> There is now an issue when editing channels for a person. The sms tag does not line up. Perhaps the best solution is to drop this. How many cell phones DON'T have SMS anymore ?

I wonder if this will be cleared up in going over the simple_form initialization.  Agreed that the SMS field could probably be removed these days.

app/assets/javascripts/datatables_defaults.js
kgf: Add it for specifics.

PR #136, opened 2016-09-12
Adding Availability model to system.

kgf:
This is a big commit for a complex area. Certainly not finished in this
area, but major functionality is here, such as tracking vacations
across events. Will also need additional tests to test when an
availability and an event overlap.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -19,5 +19,4 @@
 //= require jquery-ui-timepicker-addon
 //= require datetimepicker
 //= require jquery-ui
-//= require bootstrap
 //= require_tree .
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -17,4 +17,7 @@
 //= require jquery-ui/datepicker
 //= require jquery-ui/accordion
 //= require jquery-ui-timepicker-addon
+//= require datetimepicker
+//= require jquery-ui
+//= require bootstrap
 //= require_tree .
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -1,7 +1,7 @@
 class EventsController < ApplicationController
   before_filter :authenticate_user!
   load_and_authorize_resource
-  
+
   def schedule
     @event = Event.find(params[:id])
     @person = Person.find(params[:person_id])
@@ -28,6 +28,9 @@ class EventsController < ApplicationController
   def show
     @event = Event.find(params[:id])
     @page_title = @event.title
+    @eligible = Person.all # In the future, this will need to honor department
+    @responded = ( @event.responses ).map { |e| e.person }
+    @unknown = @eligible - @responded
     respond_to do |format|
       format.html # show.html.erb
       format.json { render json: @event }
@@ -56,7 +59,7 @@ class EventsController < ApplicationController
   # POST /events.json
   def create
     @event = Event.new(params[:event])
-    
+
     respond_to do |format|
       if @event.save
         format.html { redirect_to @event, notice: 'Event was successfully created.' }
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -28,9 +28,8 @@ class EventsController < ApplicationController
   def show
     @event = Event.find(params[:id])
     @page_title = @event.title
-    @eligible = Person.all # In the future, this will need to honor department
-    @responded = ( @event.responses ).map { |e| e.person }
-    @unknown = @eligible - @responded
+    @respondents =  @event.respondents
+    @unresponsive_people = @event.unresponsive_people
     respond_to do |format|
       format.html # show.html.erb
       format.json { render json: @event }
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -16,7 +16,6 @@ class Event < ActiveRecord::Base
   has_many :activities, as: :loggable
 
   has_many :timecards
-  has_many :availabilities
   has_many :people, :through => :timecards
 
   accepts_nested_attributes_for :timecards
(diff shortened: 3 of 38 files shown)


app/assets/javascripts/application.js
```diff
@@ -17,4 +17,7 @@
 //= require jquery-ui/datepicker
 //= require jquery-ui/accordion
 //= require jquery-ui-timepicker-addon
+//= require datetimepicker
+//= require jquery-ui
+//= require bootstrap
```
david135: According to [Bootstrap](https://github.com/twbs/bootstrap-sass#a-ruby-on-rails) (toward the bottom of that section):

> bootstrap-sprockets and bootstrap should not both be included in application.js.
> 
> bootstrap-sprockets provides individual Bootstrap Javascript files (alert.js or dropdown.js, for example), while bootstrap provides a concatenated file containing all Bootstrap Javascripts.

They [link](https://github.com/twbs/bootstrap-sass/issues/829#issuecomment-75153827) to an issue thread explaining it.

kgf: Awesome catch, thanks. I'll remove that line.

app/views/availabilities/show.html.erb
```diff
@@ -0,0 +1,32 @@
+<p id="notice"><%= notice %></p>
+
+<p>
+  <b>Person:</b>
+  <%= link_to @availability.person.fullname, @availability.person %>
+</p>
+
+<% if @availability.event.present? %>
+  <p><b>Event:</b><%= link_to @availability.event.title, @availability.event %></p>
```
david135: We could use a space here between </b> and <%= link_to

app/views/availabilities/show.html.erb
```diff
@@ -0,0 +1,32 @@
...
+<% end %>
+
+<p>
+  <strong>Start time:</strong>
+  <%= @availability.start_time %>
+</p>
+
+<p>
+  <strong>End time:</strong>
+  <%= @availability.end_time %>
```
david135, line 15: Is there a way to get the javascript to automatically bring the end time to no earlier than the start time?

app/views/people/show.html.erb
```diff
@@ -2,6 +2,7 @@
   <%= link_to 'Return to Listing', :back, class: 'btn btn-primary', style: 'width: 100%' %><br/><br/>
   <%= link_to 'Edit Person', edit_person_path(@person), class: 'btn btn-primary', style: 'width: 100%' if can? :edit, @person %><br/><br/>
   <%= link_to 'New Certification', new_person_cert_path(@person), class: 'btn btn-primary', style: 'width: 100%' if can? :create, Cert %><br/><br/>
+  <%= link_to 'New Availability', new_person_availability_path(@person), :class => 'btn btn-primary', style: 'width: 100%' if can? :create, Availability  %>
```
david135, line 5: 'New Availability' or 'Create Availability' as in events/show.html.erb?

app/views/availabilities/show.html.erb
kgf: That is a good idea, and may be handled partially by the calendar library. I will create an issue, as I don't want to hold up this pr for that.

app/views/people/show.html.erb
kgf: We should move towards new, it's shorter

app/views/availabilities/show.html.erb
kgf: Issue created #140

app/controllers/events_controller.rb
```diff
@@ -28,6 +28,9 @@ def index
   def show
     @event = Event.find(params[:id])
     @page_title = @event.title
+    @eligible = Person.all # In the future, this will need to honor department
```
kgf: I'm going to move this into the class

app/models/event.rb
```diff
@@ -2,53 +2,83 @@ class Event < ActiveRecord::Base
...
+
+  def unresponsive_people
+    eligible_people - respondents
+  end
+
   def manhours
     self.timecards.sum('actual_duration')
   end
-  def unknown_people 
+  def unknown_people
```
kgf: This is replaced by unresonsive people above

db/migrate/20160901210947_create_availabilities.rb
```diff
@@ -0,0 +1,14 @@
...
+    create_table :availabilities do |t|
+      t.datetime :start_time
+      t.datetime :end_time
+      t.string :status
+      t.text :description
+
+      t.timestamps null: false
+    end
+    add_reference :availabilities, :person, index: true
+    add_reference :availabilities, :event, index: true
```
kgf: I'm going to remove this, not certain it is needed.

app/models/event.rb
```diff
@@ -2,53 +2,83 @@ class Event < ActiveRecord::Base
...
 
   attr_accessible :title, :description, :category, :course_id, :duration, :start_time, :end_time, :instructor, :location, :status, :timecard_ids, :person_ids, :comments
- 
-  def end_date_cannot_be_before_start
-    if ((!end_time.blank?) and (!start_time.blank?)) and end_time < start_time
-      errors.add(:end_time, "must be after the start, unless you are the Doctor")
-    end
-  end
+
   def end_date_cannot_be_before_start
```
david135: Could this be changed to `end_time_cannot_be_before_start` (here and in the `validate` line)?  I'd find it more readable to have this lower in the file, among the other methods, probably in the `private` section.

kgf: For some reason, I was thinking it had to be defined before it could be called. Looking [at Ruby Guides](http://guides.rubyonrails.org/active_record_validations.html#custom-methods) shows that is not the case.

PR #137, opened 2016-09-12
Link cleanup

david135:
Aiming for consistency in views, moving navigational links to buttons in the sidebar. Using helper functions to avoid cluttering the views with explicit setting of classes and styles.

### DIFF
--- a/app/views/inspections/new.html.erb
+++ b/app/views/inspections/new.html.erb
@@ -1,5 +1,6 @@
-<h1>New inspection</h1>
+<% content_for :sidebar do %>
+  <h1>New inspection</h1>
+  <%= sidebar_button_link 'Back', inspections_path %>
+<% end %>
 
 <%= render 'form' %>
-
-<%= link_to 'Back', inspections_path %>
--- a/app/views/locations/show.html.erb
+++ b/app/views/locations/show.html.erb
@@ -1,8 +1,11 @@
-<p id="notice"><%= notice %></p>
+<% content_for :sidebar do %>
+  <%= sidebar_button_link 'Edit', edit_location_path(@location) %>
+  <%= sidebar_button_link 'Back', locations_path %>
+<% end %>
 
 <p>
   <b>Department:</b>
-  <%= @location.department.name %>
+  <%= @location.department&.name %>
 </p>
 
 <p>
@@ -39,7 +42,3 @@
   <b>Lon:</b>
   <%= @location.lon %>
 </p>
-
-
-<%= link_to 'Edit', edit_location_path(@location) %> |
-<%= link_to 'Back', locations_path %>
--- a/app/views/people/edit.html.erb
+++ b/app/views/people/edit.html.erb
@@ -1,6 +1,7 @@
 <% content_for :sidebar do %>
   <h1>Editing Person</h1>
-  <%= link_to 'Cancel', :back, class: 'btn btn-primary', style: 'width: 100%' %><br/><br/>
+  <%= sidebar_button_link 'Cancel', :back %>
+  <br>
   <%= image_tag @person.portrait_url if @person.portrait.file.present? %>
 <% end %>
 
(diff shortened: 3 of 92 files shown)


app/views/inspections/new.html.erb
```diff
@@ -1,5 +1,6 @@
-<h1>New inspection</h1>
+<% content_for :sidebar do %>
+  <h1>New inspection</h1>
```
kgf, line 2: Not sure the heading should be inside the `:sidebar`

kgf: This is a question in a few places.

kgf: NM, I think that is the norm

app/views/repairs/_repairs_table.html.erb
```diff
@@ -18,8 +18,7 @@
         <td><%= repair.repairer_name %></td>
         <td><%= number_to_currency(repair.cost) %></td>
         <td><%= repair.service_date %></td>
-        <td><%= link_to 'Edit', edit_repair_path(repair),
-                class: 'btn btn-xs btn-primary' unless cannot? :edit, repair %></td>
+        <td><%= table_button_link 'Edit', edit_repair_path(repair) unless cannot? :edit, repair %></td>
```
kgf, line 21: In retrospect, I wish I had not used `unless`
`link if can? :edit repair`

app/views/people/edit.html.erb
```diff
@@ -1,6 +1,7 @@
 <% content_for :sidebar do %>
   <h1>Editing Person</h1>
-  <%= link_to 'Cancel', :back, class: 'btn btn-primary', style: 'width: 100%' %><br/><br/>
+  <%= sidebar_button_link 'Cancel', :back %>
```
kgf, line 3: This one is different from all the rest.

app/views/people/show.html.erb
```diff
@@ -1,7 +1,8 @@
 <% content_for :sidebar do %>
-  <%= link_to 'Return to Listing', :back, class: 'btn btn-primary', style: 'width: 100%' %><br/><br/>
-  <%= link_to 'Edit Person', edit_person_path(@person), class: 'btn btn-primary', style: 'width: 100%' if can? :edit, @person %><br/><br/>
-  <%= link_to 'New Certification', new_person_cert_path(@person), class: 'btn btn-primary', style: 'width: 100%' if can? :create, Cert %><br/><br/>
+  <%= sidebar_button_link 'Return to Listing', :back %>
```
kgf, line 2: This should be an absolute rather than relative path.
`people_path`

david135: If you mean `:back` [my replacement for `session[return_to]`, which caused the problems I mentioned in #123], I think that may lead to problems, since the user may reach this page from many menu paths: People->[Police, Cert, Police Active, Other Active, Everybody, etc.]. As it is now, Other Active->Show->Return to Listing brings the user back to Other Active.  Changing it to people_path would return to that page regardless of the original page.

kgf: I see, OK

app/views/people/edit.html.erb
david135: Yeah, I think I found this one in my next PR.  I'll see if I can grab that change into this one for the final version.

app/views/locations/show.html.erb
```diff
@@ -1,8 +1,11 @@
-<p id="notice"><%= notice %></p>
+<% content_for :sidebar do %>
+  <%= sidebar_button_link 'Edit', edit_location_path(@location) %>
+  <%= sidebar_button_link 'Back', locations_path %>
+<% end %>
 
 <p>
   <b>Department:</b>
-  <%= @location.department.name %>
+  <%= @location.department&.name %>
```
kgf, line 8: What is the reason for the ampersand?

kgf:
You removed a lot of"notice" paragraphs. Is there one present at a higher level?

app/views/locations/show.html.erb
david135: Defensive coding (Ruby 2.3's new Safe Navigation Operator).  The view blew up when I happened to test it on a location whose department wasn't set properly.  Happy to remove if we're going to ensure that the value is set. I think to_s would also work, but somehow the view needs to be protected.

david135:
All the notice displays display in the sidebar from application.html.erb (going back for a long time).  When we also had these notice paragraphs, the messages showed twice.

app/views/locations/show.html.erb
kgf: I didn't know about that, I'll look it up.

Issue #134, opened 2016-09-10
Item Inspections can't be saved


kgf:
When creating an inspection, you can't save it, results in a 500. I believe this is related to "Forbidden Attributes".
The ideal PR correcting this would include a test that exercises this code.

david135:
Hopefully addressed in PR #138.

PR #138, opened 2016-09-12
Item inspections

david135:
Addressing #134 and continuing improvements. Probably best follows PR #137.
Of note:
1. This is complicated by the ability to add an inspection both from the inspections index page and from an item. (I was able to imagine a use case, by an administrator who wants to schedule inspections, then choosing the items.)  It could be simplified in several ways if we only added inspections from an item.
2. I added inspection_date and item validations to inspection.rb
3. Browser-side validations of forms is now enabled, in order to enforce the validations cleanly.
`config.browser_validations = true` (which is now the default, but we had it as false)
4. A couple of tests included.

Looking forward to your thoughts.


kgf:
Inspections will be initiated from the item show. In real life, this might be a qr code directly to the item. They will not be from an inspection index path. I prefer not to have two paths, as that will make training harder. "Go to the item, then click inspection."

app/assets/javascripts/inspections.js.coffee
```diff
@@ -0,0 +1,20 @@
+jQuery ->
```
kgf, line 1: Could you explain what this is doing? Thanks

kgf: I mean this block of js, not that one line.

app/controllers/inspections_controller.rb
```diff
@@ -27,7 +27,11 @@ def show
   # GET /inspections/new
   # GET /inspections/new.json
   def new
-    @inspection = Inspection.new
+    if @item
+      @inspection = @item.inspections.build
+    else
+      @inspection = Inspection.new
```
kgf: If we removed the inspection/new path from the routes, this could be simplified. We would always have an item.

app/controllers/inspections_controller.rb
```diff
@@ -37,14 +41,16 @@ def new
...
 
   # POST /inspections
   # POST /inspections.json
   def create
-    @inspection = Inspection.new(params[:inspection])
+    # See below in set_item. if we didn't receive the item from the beginning,
+    # we get it here from the form.
+    @item ||= Item.find(params[:inspection][:item_id])
 
+    @inspection = @item.inspections.build(inspection_params)
```
kgf, line 22: I like scoping it to the item

app/controllers/inspections_controller.rb
```diff
@@ -75,12 +79,30 @@ def update
...
   end
+
+  private
+    def set_item
+      # params[:item_id] may not be set, because we can create a new
+      # inspection from the inspections index page, adding its item in
+      # the form.
+      # The two routes are:
+      # new_item_inspection_path(@item)
+      # new_inspection_path
```
kgf: Remove this path

app/helpers/application_helper.rb
```diff
@@ -11,6 +11,74 @@ def nav_link(link_text, link_path, http_method=nil)
...
+  # Automatically adds class and/or delete confirmation (if absent),
+  # before passing them on to link_to.
+  # Uses add_class_and_confirmation, which contains the logic.
+  # Uses btn-block (which spans the width of the parent element) for the sidebar.
+  # Examples:
+  # <%= sidebar_button_link 'Edit Person', edit_person_path(@person) %>
+  # <%= sidebar_button_link 'Edit Person', edit_person_path(@person), class: 'btn btn-block btn-warning' %>
+  # <%= sidebar_button_link 'Destroy Person', @person, method: :delete %>
+  # <%= sidebar_button_link 'Destroy Person', @person, method: :delete, class: 'btn btn-block btn-success' %>
+  def sidebar_button_link(name = nil, options = nil, html_options = nil, &block)
```
kgf, line 35: What if we called something else, like context nav, rather than trying it to the location, which I would like to change anyway.

david135:
> Inspections will be initiated from the item show. In real life, this might be a qr code directly to the item. They will not be from an inspection index path. I prefer not to have two paths, as that will make training harder. "Go to the item, then click inspection."

That will simplify the code as well, though I think I now have an implementation of the dual path that looks better than my first attempt.

app/helpers/application_helper.rb
kgf: I realize now that this is actually part of the other PR

app/assets/javascripts/inspections.js.coffee
david135: Good catch.  This should not be in both inspections.js.coffee and items.js.coffee.

david135:
1. I've switched to using the `datetimepicker/datepicker` that was installed in the availabilities PR and moved configuration to `app.js.coffee`.
2. There is some complication with editing a form whose `date` or `datetime` field is set by the `datepicker`/`datetimepicker`.  If the value is not formatted properly, problems occur.  I've fixed this for Inspection and Repair, but it should be fixed everywhere.  I've addressed this by setting the value in the form using helper functions, `format_datetime_value` or `format_date_value`.  Try editing an `availability` to see this complication before the fix.
3. In the routes, I've removed some unused routes.  Because inspections are only created for particular items, create and new must be reached through an item, while others go directly.  See [here](http://guides.rubyonrails.org/routing.html#limits-to-nesting) for the motivation.
4. While this PR was aimed at Inspection, I've made some small changes to Repair that were inspired by their similarities.
5. Inspection uses `datetime` in the db while Repair uses `date`.  Is this the intention? I could imagine both an inspection and a repair making sense as scheduled simply for a day or scheduled for a particular time on a particular date.
6. There are more improvements to Repair that could be made in a future PR.
7. Small fixes to `departments_controller.rb`, `app/views/items/show.html.erb`, `app/views/events/show.html.erb`

kgf:
On the new inspection form, the field for the date time of the inspection is not big enough.
screen shot 2016-09-21 at 6 59 23 pm

kgf:
Could you please add an additional column to the datatable on the item that lists inspections to include  the comments. `truncate` might be a good thing here.

app/views/inspections/index.html.erb
```diff
@@ -11,14 +16,17 @@
   <tbody>
     <% @inspections.each do |inspection| %>
       <tr>
+        <td><%= inspection.item&.name %></td>
```
kgf: Please make this a link to the item

app/views/inspections/index.html.erb
```diff
@@ -11,14 +16,17 @@
...
+        <td><%= format_datetime_value(inspection.inspection_date) %></td>
```
kgf: Please make this a link to the inspection.

app/views/inspections/index.html.erb
```diff
@@ -11,14 +16,17 @@
...
+        <td>
+          <span class="label <%= inspection_label(inspection) %>">
+            <%= inspection.status %>
+          </span>
+        </td>
         <td><%= table_button_link 'Show', inspection %></td>
```
kgf: You can remove this, no longer needed

app/views/inspections/index.html.erb
```diff
@@ -11,14 +16,17 @@
...
         <td><%= table_button_link 'Edit', edit_inspection_path(inspection) %></td>
         <td><%= table_button_link 'Destroy', inspection, method: :delete %></td>
```
kgf: You should permission this and also the edit action. Actually, you might be able to add the permissions to the helper, which would be **a good thing**. No permission, the helper returns an empty string

app/views/inspections/index.html.erb
```diff
@@ -11,14 +16,17 @@
   <tbody>
     <% @inspections.each do |inspection| %>
       <tr>
+        <td><%= inspection.item&.name %></td>
+        <td><%= format_datetime_value(inspection.inspection_date) %></td>
+        <td>
+          <span class="label <%= inspection_label(inspection) %>">
+            <%= inspection.status %>
+          </span>
+        </td>
```
kgf: Please add a truncated version of the comments here as well.

app/views/inspections/index.html.erb
david135: I've added permissions.  I think the helper(s) will need pretty careful design to do this cleanly.

app/views/inspections/show.html.erb
```diff
@@ -3,7 +3,7 @@
   <h3>for <%= @inspection.item.name %></h3>
   <%= sidebar_button_link 'Return to Item', @inspection.item %>
   <%= sidebar_button_link 'Return to Inspections List', inspections_path %>
-  <%= sidebar_button_link 'Edit Inspection', edit_inspection_path(@inspection) unless cannot? :edit, @inspection %>
+  <%= sidebar_button_link 'Edit Inspection', edit_inspection_path(@inspection) if can? :edit, @inspection %>
```
kgf, line 6: Thanks for the switch

PR #141, opened 2016-09-17
86 Filter locations based on department on the item create and edit p…

pragyajalan:
…ages

```
modified:   app/assets/javascripts/items.js.coffee
modified:   app/views/items/_form.html.erb
```

Address the filtering https://github.com/ReadyResponder/ReadyResponder/issues/86

### DIFF
--- a/app/assets/javascripts/items.js.coffee
+++ b/app/assets/javascripts/items.js.coffee
@@ -16,3 +16,13 @@ jQuery ->
                   { responsivePriority: 5, targets: 6 },
                   { orderable: false, targets: 9 },
                 ]
+
+  locations = $('#item_location_id').html()
+  $('#item_department_id').change ->
+    department = $('#item_department_id :selected').text()
+    espaced_department = department.replace(/([ #;&,.+*~\':"!^$[\]()=>|\/@])/g, '\\$1')
+    options = $(locations).filter("optgroup[label='#{espaced_department}']").html()
+    if options
+      $('#item_location_id').html(options)
+    else
+      $('#item_location_id').parent().empty()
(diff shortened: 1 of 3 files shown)


david135 (commented):
The JS could use some commenting.

`db/seeds.rb` would benefit from having locations associated with departments so that this can be tested more easily.

The Location box disappears if Department is cleared after having been set (= unset). It does not come back after re-setting the Department.

Perhaps we would do better to restore the Location list to an unfiltered state if Department is cleared.

(Another possibility with different behavior: simply set the department and location at the same time from the list of locations organized under department. This could work pretty smoothly using the grouped_select that you've implemented. @kgf ?)

kgf:
Sadly, I have a bigger concern regarding this PR. Let's say I have Department A and Department B. I have two locations. Location A1 is part of department A, and Location B2 is part of Department B. I choose Department A from the first drop down. I can _still_ choose Location 2 from the location drop down. This leaves me with an invalid situation, where the location does not belong to the department.

I did not write a good issue. A good user story for this would be:
A user starts a new item. There are no locations available until they choose a department. (Maybe the dropdown is disabled) Once they have chosen a department, there are locations to chose from. Only the locations for that department are shown. If they change their department selection, the choices should change too.

pragyajalan:
@kgf Im not sure I understand your concern. 

In your example If you choose Department A, you will only see options for that Department A. If you change that to Department B, the options should only show the options for Department B. Isnt that what should happen?

kgf:
That was not what I recall seeing. I'll setup a better example for you.

Thanks !

kgf:
@pragyajalan I did check out this branch again, not sure what happened the first time. It is now doing for me what you described. Great !

But.....
If I choose a department, the location is filled in with the first choice. I can not remove it. When changing the department, I would like the location to remain blank, so it can't be accidentally set to an incorrect value.

Thanks for your help.

pragyajalan:
@kgf Just pushed the change to add an empty option when choosing the location for a department

Issue #58, opened 2015-12-19
Org chart report should include first initial


kgf:
Then you could better tell one person from another.

PR #143, opened 2016-09-23
Signin sheet can now enumerate division1 values

kgf:
This means that when we get to the division 1 choices being a setting,
rather than a class constant, the report will still work.

This also addresses #58

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -6,6 +6,7 @@ class PeopleController < ApplicationController
   def signin
     #This is the sign-in sheet, not anything about authentication
     @people = Person.active
+    @department = "Police"
     @page_title = "Sign-in"
     render :layout => "print_signin"
   end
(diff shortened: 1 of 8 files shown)


app/controllers/people_controller.rb
```diff
@@ -6,6 +6,7 @@ class PeopleController < ApplicationController
   def signin
     #This is the sign-in sheet, not anything about authentication
     @people = Person.active
+    @department = "Police"
```
kgf, line 9: This is _not_ a long term solution, just a step on the way

david135:
Works for me. It would possibly benefit from making sure initials are capitalized, etc., but hopefully our validators will eventually handle that. In the long run, I'd look for less confusing names for 'signin' ('signin_sheet', perhaps?), `division1`, and `division2`.  The last two are a bit confusing given that departments may also have a "Division 1" and "Division 2".

I know that isn't the purpose of this PR, but since you mention Issue #58, the Org Chart seems a bit brittle.  I don't have terribly realistic data in my db, but I was able to produce this:
broken_orgchart

Issue #144, opened 2016-09-23
Items show form has gaps


kgf:
When you look at this form, you see two tables, repairs and inspections. There is a large gap between the two. On smaller screens you may not even realize the inspections are there.

Ideally this should be much smaller. Perhaps the repairs table could grow to the needed size, up to a small maximum.

david135:
This is because of the default setting of `scrollY` for datatables, set to `50vh` (IIRC, half of the vertical window's height).  A height needs to be set, so that we can use the scrollY feature and keep the heading row visible, but 50vh is simply a rough guess for what might work for many tables.  The tables on this page (and possibly others) will need a different value.

david135:
Addressed in #146.

Issue #142, opened 2016-09-22
A person's date of birth is not validated, and it is not saved if entered according to the placeholder


david135:
A person's date of birth is not validated, and it is not saved if entered according to the placeholder. One can enter a date of birth in the year-first order, and it will save.

Issue #140, opened 2016-09-15
Making sure dates are in the correct order


kgf:
On several forms, we enter start and end times. There should be validations on all these models to require that `end_time` is after `start_time`.  What would be much better would be to have validations in the form, using Javascript, to catch these errors before they get to the backend.

The ideal PR would verify that all models have validations, that there is JS to help the user before form submission, and tests to test it !

david135:
I've got the validations and tests taken care of in my next PR. I don't yet have a plan for JS.

kgf:
I would accept that PR, and make a new issue for the JS.

david135:
PR #146 should help with this.

Issue #133, opened 2016-09-09
Change Items View to better reflect reality


kgf:
We'll need to change the item view. We should remove the `serial X` columns, as they don't really apply anymore. Also, I'd like to remove the `Available` status column, and just colorize the row itself. The concern I have is if it's printed on a b/w printer, that info would be lost. 
https://datatables.net//forums/discussion/34882

david135:
Table columns removed in PR #146, but will we want to remove these fields from the db?

kgf:
Yes, a migration should be created to remove these fields.

david135:
Resolved with #185.

PR #146, opened 2016-09-23
validate chronology ; improve some views and specs

david135:
Addresses #140, #142, #133, and #144.  Factored out of the work I'd begun on tasks.

The most interesting piece is in `config/initializers/validators.rb`, where the validator `validates_chronology` is defined, to be available in all of our models.  It is used as follows:
`validates_chronology :start_time, :end_time`

I've used it in models that seem to need it, and confirmed that they passes existing tests for chronology, and added tests to models that hadn't tested this.

`item.rb` has a couple of fields that may or may not need this validation.  They are there, but commented, and tests have not been added.

I've also fixed datepicker/datetimepicker's functionality in a number of forms (the problem manifested as corruption of a field's contents when editing a form with a saved date/datetime value) .

I've fixed a few cosmetic issues I've discovered.

### DIFF
--- a/app/models/item.rb
+++ b/app/models/item.rb
@@ -8,6 +8,9 @@ class Item < ActiveRecord::Base
                   :department_id, :resource_type_id, :item_type_id,
                   :unique_ids_attributes
 
+  # validates_chronology :purchase_date, :sell_date     # ? - if so, needs a test
+  # validates_chronology :grantstart, :grantexpiration  # ? - if so, needs a test
+
   mount_uploader :item_image, ItemImageUploader
   validates_numericality_of :value
   belongs_to :owner, class_name: 'Person', inverse_of: :items
--- a/app/models/timecard.rb
+++ b/app/models/timecard.rb
@@ -1,7 +1,7 @@
 class Timecard < ActiveRecord::Base
   before_save :pull_defaults_from_event
   before_save :calc_durations
-  attr_accessible :intention, :intended_start_time, :intended_end_time, 
+  attr_accessible :intention, :intended_start_time, :intended_end_time,
                   :outcome, :actual_start_time, :actual_end_time, :event_id, :person_id, :category, :description
 
   belongs_to :person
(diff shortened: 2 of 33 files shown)


app/models/item.rb
```diff
@@ -8,6 +8,9 @@ class Item < ActiveRecord::Base
                   :department_id, :resource_type_id, :item_type_id,
                   :unique_ids_attributes
 
+  # validates_chronology :purchase_date, :sell_date     # ? - if so, needs a test
+  # validates_chronology :grantstart, :grantexpiration  # ? - if so, needs a test
```
kgf, line 12: Grants will be a separate model, so the validations will go there

app/models/timecard.rb
```diff
@@ -10,27 +10,11 @@ class Timecard < ActiveRecord::Base
...
-      errors.add(:intended_end_time, "must be after the start, unless you are the Doctor")
-    end
-  end
-
-  def actual_end_date_cannot_be_before_actual_start
-    if ((!actual_end_time.blank?) and (!actual_start_time.blank?)) and actual_end_time < actual_start_time
-      errors.add :actual_end_time, "must be after the start, unless you are the Doctor"
-    end
-  end
+  validates_presence_of :person_id, :event_id
```
kgf, line 13: A future PR will need to remove the link to events. A Timecard represents a person working for a period of time, which could include multiple events.

app/views/availabilities/_form.html.erb
```diff
@@ -3,14 +3,11 @@
...
-                :collection => (Availability::STATUS - ["Cancelled"]) %>
+    <%= f.input :status, :as => :select, :collection => (Availability::STATUS - ["Cancelled"]) %>
     <%= f.input :description %>
 
-    <%= f.input :start_time, :as => :string,
-                :input_html => { :class => 'datetimepicker', :size => 11 }  %>
-    <%= f.input :end_time, :as => :string,
-                :input_html => { :class => 'datetimepicker', :size => 11 }  %>
+    <%= f.input :start_time, as: :string, input_html: { class: 'datetimepicker', size: 11, value: format_datetime_value(f.object.start_time) } %>
+    <%= f.input :end_time, as: :string, input_html: { class: 'datetimepicker', size: 11, value: format_datetime_value(f.object.end_time) } %>
```
kgf: These are very long lines, which do not fit into the default github UI. Please break them across two lines, similar to the originals. On the plus side, yours are more accurate !

app/views/inspections/_form.html.erb
```diff
@@ -2,8 +2,7 @@
   <%= f.error_notification %>
 
   <div class="form-inputs">
-    <%= f.input :inspection_date, as: :string, input_html: { class: 'datetimepicker', size: 11,
-                                                             value: format_datetime_value(f.object.inspection_date) } %>
+    <%= f.input :inspection_date, as: :string, input_html: { class: 'datetimepicker', size: 11, value: format_datetime_value(f.object.inspection_date) } %>
```
kgf: Not even sure what changed here, other than all on one line.

spec/features/inspection_display_spec.rb
```diff
@@ -13,8 +13,8 @@
     click_on 'Sign in'
   end
   describe "when not logged in" do
-    let(:an_item)  { FactoryGirl.create(:item) }
-    let(:an_inspection) { FactoryGirl.create(:inspection, item: an_item) }
```
kgf, line 17: Why are you removing FactoryGirl ?

spec/models/cert_spec.rb
```diff
@@ -4,11 +4,17 @@
...
-  it 'sets the issued date to today' do
+  it 'sets the issued date to today' do # This is set in the factory
     expect(subject.issued_date).to be == Date.today
   end
 
   it 'sets the expiration date properly' do
     expect(subject.expiration_date).to be == Date.today + evoc_course.term.to_i.months
   end
+
+  it "requires expiration_date to be after issued_date" do #chronology
```
kgf, line 15: I like this, new tests !

kgf:
Thanks for the new tests !

spec/features/inspection_display_spec.rb
david135: I'd realized that they were systematically removed in b975d6f2ffb2ff273b94c0f5221bca28747307fd . If that was the standard, I'd erred in putting them in myself.

PR #149, opened 2016-09-27
display_cleanup - followup to previous PRs and preliminary to upcoming PR

david135:
display_cleanup - followup to previous PRs and preliminary to upcoming PR

Several small improvements and cleanups:
1) Fix datatable initialization in `items.js.coffee` which broke after removing columns.
2) Improve label helpers; apply to code base (e.g., `inspection_status_label`).
3) Highlight a row based on status (`success_class_if_match`), e.g.,
`<tr <%== success_class_if_match(availability.status, 'Available') %>>...</tr>`
4) Fix a few classes for datatables.
5) Improve styling of `people_form.html.erb` (somewhat broken since Bootstrap 3).
6) Improve error message for the `validates_chronology` helper.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -92,11 +92,13 @@ module ApplicationHelper
   # label.  For anything more precise, a custom helper should be
   # used (as in items_helper.rb)
   def success_or_warning_label(str, success_str)
-    if str == success_str
-      return 'label-success'
-    else
-      return 'label-warning'
-    end
+    content_tag(:span, str, class: success_or_warning_label_class(str, success_str))
+  end
+
+  # This is a general helper to provide a simple success tag when appropriate.
+  # This should be called with raw or with <%==
+  def success_class_if_match(str, success_str)
+    return 'class="success"' if str == success_str
   end
 
   # This is for providing a date for a form in an order that Rails can understand.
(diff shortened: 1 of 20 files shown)


app/helpers/application_helper.rb
```diff
@@ -92,11 +92,13 @@ def link_to_add_fields(name, f, association)
   # label.  For anything more precise, a custom helper should be
   # used (as in items_helper.rb)
   def success_or_warning_label(str, success_str)
```
kgf, line 94: Should this be `sucess_or_warning_tag`, as that's what it returns ?

app/views/courses/show.html.erb
```diff
@@ -6,10 +6,7 @@
 <% end %>
 
 <p>
-  <b>Status:</b>
-  <span class="label <%= success_or_warning_label(@course.status, 'Active') %>">
-    <%= @course.status %>
-  </span>
+  <b>Status:</b> <%= success_or_warning_label(@course.status, 'Active') %>
```
kgf: It looks like this would just have the label/class ? Wouldn't we still need `<%= @course.status %>`

And if you wanted to throw in a test on this....

app/helpers/application_helper.rb
david135: it returns html for:

`<span class="label label-success">str</span>`
or
`<span class="label label-warning">str</span>`
depending on whether str matches success_str, so it does actually add a Bootstrap label.

Issue #139, opened 2016-09-13
Add task model to events


kgf:
This involves a new model, Task. The model is at the bottom of this issue.

This should have a datatable on the event, listing all the tasks for that event.
It should also have an instance method for `status`, which will look down into the requirements. For now, this method can return things like ( Empty, Partially Filled, Full <Although I'm open to other terms.>)  The value of this status should affect the color of the row in the data table, and also result in a small icon in an early column of the datatable. Additionally, if you click to sort on the icon column, Empty should be at the top, then Partially Filled, then Full.

Tasks

```
belongs_to event
# These has_many relationships need to wait for these models to be created
# has_many requirements
# has_many assignments, through requirements
title
description
street
city, state, zip
latitude
longtitude

start_time
end_time

```

If you think of other fields that should be considered, please comment.

kgf:
This was met by #152

PR #152, opened 2016-09-28
add task model

david135:
adding Task model according to #139.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -119,6 +119,14 @@ module ApplicationHelper
     datetime.strftime("%Y-%m-%d %H:%M") if datetime
   end
 
+  def format_time_span(start_time, end_time, duration = nil)
+    if duration
+      "#{start_time} to #{end_time}, #{duration}h"
+    else
+      "#{start_time} to #{end_time}"
+    end
+  end
+
   private
     def success_or_warning_label_class(str, success_str)
       if str == success_str
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -95,8 +95,14 @@ module ApplicationHelper
     content_tag(:span, str, class: success_or_warning_label_class(str, success_str))
   end
 
-  # This is a general helper to provide a simple success tag when appropriate.
-  # This should be called with raw or with <%==
+  # Returns a 'data-order="value"' string to be used in a <td> tag, which can be used for sorting
+  # This helper should be called with raw or with <%==
+  def sorting_order(value)
+    return "data-order=\"#{value}\""
+  end
+
+  # This is a general helper to provide a simple success class label when appropriate.
+  # This helper should be called with raw or with <%==
   def success_class_if_match(str, success_str)
     return 'class="success"' if str == success_str
   end
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -24,7 +24,7 @@ class Task < ActiveRecord::Base
 
   def status_value
     # return a value for sorting
-    STATUS_CHOICES[status]
+    return STATUS_CHOICES[status]
   end
 
   def requirements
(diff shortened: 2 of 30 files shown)


app/models/task.rb
```diff
@@ -0,0 +1,60 @@
...
+  # has_many :assignments, through: :requirements
+
+  STATUS_CHOICES_ARRAY = ['Empty', 'Partially Filled', 'Full', 'Cancelled']
+
+  # The following provides the equivalent of:
+  # STATUS_CHOICES = { 'Empty' => 0, 'Partially Filled' => 1, 'Full' => 2, 'Cancelled' => 3 }
+  # This is used to give a sortable numeric value based on the order of the above array.
+  STATUS_CHOICES = STATUS_CHOICES_ARRAY.map.with_index { |v, i| [v, i] }.to_h
+
+  def status
```
kgf, line 19: To be clear to others who may be watching, this is only for testing. As we build out the other models needed for tracking events, this will go away.

kgf:
I would like to adjust the UI on this.
screen shot 2016-09-29 at 11 57 17 am
I think we could use more horizontal space for the data, and lose more of the labels.
So maybe 
**Category** Title at Location
(start time to end time duration)
Status (Available count in green/unavailable count in red/unknown count in yellow)
Then the tables across all columns below that.

Thoughts ?

david135:
I was aiming to make use of currently unused space.  Compare:

The previous display of an event (`development`):
event-before task addition
with the display now (`task_model`):
event-after task addition

It sounds like with your changes, we'd be down to just a few lines at the top.

Here's my draft of that (`task_model_update1` - not yet submitted):
event-task addition-update 1

I added the word personnel, to indicate the basic meaning of the numbers.  I styled the numbers as labels, because when simply coloring the numbers, one could barely detect the colors.

I put the time display into a helper function.  I envision someone improving it so that the date isn't repeated unless necessary, etc., but will leave that for another time.

I also moved the comments/created/last modified info back to the bottom (where they had been in `development`), where one will only find them if looking for them.

Is this on the way to what you have in mind?

(In general, I wonder about scarcity of horizontal space vs vertical space.  It is sometimes a burden on the user to have to scroll down, and to the extent that we can fit what is necessary on a single "page" without scrolling, that may be a worthy aim.)

kgf:
That looks MUCH better, and is very close to what I had imagined. We should be sure to color the rows in the tables the same way. ( Make everyone in the Not Responded table yellow ) This would reinforce the meaning of the colors. A tooltip on those numbers would be helpful too.
Thanks !

kgf:
How hard would it be to hide the pagination when not needed ?

david135:
It isn't as easy as a configuration option.  It sounds like we'd need a JS callback to check the number.  Are we ever going to have more than a hundred entries for any of these tables?  We could disable paging by default and only enable it for tables we think will be large.

kgf:
I would say disable the paging then. It will be unusual for us to have that many rows in a table.

app/helpers/application_helper.rb
```diff
@@ -113,6 +119,14 @@ def format_datetime_value(datetime)
     datetime.strftime("%Y-%m-%d %H:%M") if datetime
   end
 
+  def format_time_span(start_time, end_time, duration = nil)
+    if duration
```
kgf, line 123: If you have the start time and the end time, why not calculate the duration ?

spec/support/authorization_view_helper.rb
```diff
@@ -33,4 +33,38 @@ def get_basic_editor_views(model_name, things_to_look_for)
...
+    outer_model = outer_model_name.to_sym
+    nested_model = nested_model_name.to_sym
+    nested_model_controller = nested_model_name.pluralize.to_sym
+
+    it "gets the index" do
+      # I was unable to get the later attributes call to work unless factorygirl call was within the test
+      @sample_object = create(nested_model)
+      index_model_path = url_for(action: :index, controller: nested_model_controller)
+      visit index_model_path
+      expect(page).to have_content("LIMS") # This is in the nav bar
```
kgf, line 47: Not going to stop the PR for this, but we should be starting to filter these out. The static text "LIMS" is going away.

PR #157, opened 2016-09-30
improve inspections and availabilities tests

david135:
add associations to two factories
add a few more tests
use new test helper function

### DIFF
--- a/spec/factories/availabilities.rb
+++ b/spec/factories/availabilities.rb
@@ -1,5 +1,6 @@
 FactoryGirl.define do
   factory :availability do
+    association :person
     start_time Time.now
     end_time 15.hours.from_now
     status "Unavailable"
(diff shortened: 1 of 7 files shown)


spec/factories/availabilities.rb
```diff
@@ -1,5 +1,6 @@
 FactoryGirl.define do
   factory :availability do
+    association :person
```
kgf, line 3: I would prefer to actually _not_ do this. I have found that having the Factory create the associations can dramatically slow the test suite, as it's creating objects you may not need in that test. Also, it interferes with creating stubs/mocks. I accept that it can make some tests longer.
Does that make sense ?
Thanks

kgf:
@david135  would you still like to get this in, or should we close it?

kgf:
Sadly, @david135 has not had time recently to resolve this. I am going to close it in anticipation of Hacktoberfest. I will re-open it if someone requests that.

Issue #145, opened 2016-09-23
Department model needs fields for Division1 and Division2 choices


kgf:
These should be stored to make it easy to use them in the person form.

kgf:
This was addressed by #159

PR #159, opened 2016-10-03
Division fields

pcofilada:
Added new division1 and division2 fields on Department model

### DIFF
--- a/app/models/department.rb
+++ b/app/models/department.rb
@@ -1,5 +1,7 @@
 class Department < ActiveRecord::Base
-  attr_accessible :contact_id, :description, :name, :status
+  serialize :division1, Array
+  serialize :division2, Array
+  attr_accessible :contact_id, :description, :name, :status, :division1, :division2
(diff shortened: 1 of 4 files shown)


pcofilada:
@kgf Can you check this one? And let me know if there is any other changes you want to make with this.

pcofilada:
#145

kgf:
I really appreciate your help. Could you add some values to the `seeds.rb` file. A department might have  "Team 1" and "Team 2" as division1 values, and "Squad 1" and "Squad 2" as division 2 values. The strings should be serialized arrays so that they can be used as choices on the simple forms
http://stackoverflow.com/questions/19134235/how-to-add-to-a-serialized-array

pcofilada:
@kgf updated.

kgf:
Thanks. Do you know why there were so many changes to schema.rb ?

pcofilada:
Hmmm. Maybe because I'm using the latest ruby version?

PR #166, opened 2016-10-05
Removed Sqlite gem from Gemfile since it is not used

phsacramento:
- Removed Sqlite Gem from any envs


david135:
Thanks for your submission! I see that someone else beat you to it, but we appreciate your help.

Issue #163, opened 2016-10-04
Remove sqllite from Gemfile


kgf:
We are using PG in all environments

pratyushprakash:
Hi, can I work on this issue?

david135:
This issue has been claimed.  Thank you for your interest, but we wouldn't want you to duplicate work that's already being done.  Perhaps another issue will catch your interest?

david135:
Resolved in #164.  Thanks for everyone's interest!

PR #168, opened 2016-10-05
Removing sqlite gem from Gemfile

jasonsouza:
Removing sqlite gem per issue #163


david135:
Thanks for your submission! I see that someone else beat you to it, but we appreciate your help.

PR #174, opened 2016-10-05
Initial work to get environmnts and gems up 2 date

javimuu:


kgf:
I am going to close this, as the original submitter did not respond, and had not actually made any commits into this branch. I hope javimuu will return and take on one of the existing issues, but in the mean time, these PRs are clutter.

PR #176, opened 2016-10-05
Start of AJAX, but I don't like it

javimuu:
I probably will not continue with this approach, instead I'll do
checkboxes and bulk update, and ember JS later.


kgf:
I am going to close this, as the original submitter did not respond, and had not actually made any commits into this branch. I hope javimuu will return and take on the Javascript he has described in the comment, but in the mean time, these PRs are clutter.

Issue #171, opened 2016-10-05
Code of Conduct needs contact information


VanAuken:
In the Code of Conduct referenced in README,  contact information needs to be inserted in document.  Currently reads:

"contacting a project maintainer at [INSERT EMAIL ADDRESS]."

kgf:
That email address should be `conduct@readyresponder.org`.
That address is not yet hooked up. In fact the readyresponder domain is not properly configured, but this will be a step in the correct direction.

ankit01ojha:
Is the project is removed from hacktoberfest?

david135:
Hi @ankit01ojha, thanks for your interest!  This project is still taking part in Hacktoberfest, but this particular issue had been claimed already, and in fact has now been resolved with PR #177.

kgf:
Hi @ankit01ojha, thanks for your interest! We do still have Hacktoberfest issues available, if we can tempt you....

Kevin

PR #177, opened 2016-10-05
Add Contact Information 

suvojit-0x55aa:


suvojit-0x55aa:
Response to issue #171

PR #194, opened 2016-10-08
Changing all specs to require rails_helper instead spec_helper.

### LINKED ISSUE #96: Fix spec_helper vs rails_helper
The Rails generators are adding `rails_helper` to specs, but we currently have `spec_helper`. We would need to change all the references.

caironoleto:
This commit closes #96


david135:
Thanks for taking this on.  Before merging this, I'd like us to discuss making sure that only test files that rely on rails use `rails_helper`.  This is discussed a bit [here](https://www.relishapp.com/rspec/rspec-rails/docs/upgrade#default-helper-files). Thoughts?

caironoleto:
@david135 Thank you for the observation. I changed the spec_helper and rails_helper following the upgrade guide. But all specs are related with rails.

david135:
Thank you!  I see you've gone on updating the specs (to current RSpec defaults?).  It might help reviewers if you could write a few lines (here?) summarizing the changes.

I found [this](http://stackoverflow.com/questions/24747448/rspec-describe-vs-describe) discussion of `describe` vs `RSpec.describe`, which might help others in understanding the changes.

I see that we're no longer requiring `database_cleaner` in `spec_helper.rb` (nor in `rails_helper.rb`).  If we are using it, perhaps we'll want to update its configuration in `spec/support/database_cleaner.rb`. @caironoleto, this need not be your task unless you want it, but I wanted the other reviewers to be aware. [Here](https://github.com/DatabaseCleaner/database_cleaner) is the current documentation.

Are there any changes that team members should be aware of in writing new tests?

In going through the tests, did you notice any improvements that could/should be made?  If so, we'd be grateful if you opened an issue.

I see that regarding `examples.txt`, they write:

> We recommend  you configure your source control system to ignore this file.

Could you please add this file to `.gitignore` and not commit it?

Thanks for your work in helping bring us up to current standard practice!

david135:
I'd also like to see what others think about the profiling by default.  It's good to have available, but it gives us a large amount of output to look at with each run (20+ lines).

For those trying it out, `rspec spec --no-profile` will disable profiling if we have `config.profile_examples` set, and similarly, `rspec spec --profile` will enable it if we don't configure it to default to profile. See [here](https://www.relishapp.com/rspec/rspec-core/docs/configuration/profile-examples) for the documentation.

kgf:
I do kinda like the profiling info. We still have the last line as `348 examples, 0 failures, 8 pending`, so the extra info is not distracting to me.
I'm in favor of it. 
Any other opinions ?

kgf:
I think this PR should include an entry into `.gitignore` for `spec/examples.txt`

caironoleto:
@kgf @david135 thank you for the feedback.

david135 (approved):
Thanks very much!

kgf:
@caironoleto Thanks for your help !

PR #203, opened 2016-10-16
use sign_in_as helper ; small spec cleanups

david135:
Added a helper function and made some small cleanups.

In feature specs, instead of:

```
before (:each) do
    somebody = create(:user)
    r = create(:role, name: 'Reader')
    somebody.roles << r
    visit new_user_session_path
    fill_in 'user_email', :with => somebody.email
    fill_in 'user_password', :with => somebody.password
    click_on 'Sign in'
end
```

use:

```
before (:each) { sign_in_as('Reader') }
```

### DIFF
--- a/spec/features/authorization_editor_spec.rb
+++ b/spec/features/authorization_editor_spec.rb
@@ -1,16 +1,13 @@
 require 'rails_helper'
-      #save_and_open_page
+
 RSpec.describe "a user" do
+  before(:each) do
+    @person = create(:person)
+  end
+
   describe "in the reader role" do
-    before (:each) do
-      @person = create(:person)
-      somebody = create(:user)
-      somebody.roles << create(:role, name: 'Reader')
-      visit new_user_session_path
-      fill_in 'user_email', :with => somebody.email
-      fill_in 'user_password', :with => somebody.password
-      click_on 'Sign in'
-    end
+    before (:each) { sign_in_as('Reader') }
+
     it "cannot edit people" do
       visit edit_person_path(@person)
       expect(page).to have_content("Access Denied")
@@ -35,16 +32,10 @@ RSpec.describe "a user" do
       expect(page).not_to have_content(@person_inactive.lastname)
     end
   end
+
   describe "in the editor role" do
-    before (:each) do
-      @person = create(:person)
-      somebody = create(:user)
-      somebody.roles << create(:role, name: 'Editor')
-      visit new_user_session_path
-      fill_in 'user_email', :with => somebody.email
-      fill_in 'user_password', :with => somebody.password
-      click_on 'Sign in'
-    end
+    before (:each) { sign_in_as('Editor') }
+
     it "can edit people" do
       visit people_path
       expect(page).to have_content('Edit') #Need to scope this, or it will fail on Edith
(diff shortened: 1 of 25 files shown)


spec/features/authorization_editor_spec.rb
```diff
@@ -1,16 +1,13 @@
...
-    before (:each) do
-      @person = create(:person)
-      somebody = create(:user)
-      somebody.roles << create(:role, name: 'Reader')
-      visit new_user_session_path
-      fill_in 'user_email', :with => somebody.email
-      fill_in 'user_password', :with => somebody.password
-      click_on 'Sign in'
-    end
+    before (:each) { sign_in_as('Reader') }
```
kgf, line 9: Is having two different `before :each` blocks OK ?

spec/features/authorization_guest_spec.rb
```diff
@@ -1,25 +1,24 @@
...
 RSpec.describe "a user" do
   describe "without a role" do
     before (:each) do
       @person = create(:person)
-      somebody = create(:user)
-      visit new_user_session_path
-      fill_in 'user_email', :with => somebody.email
-      fill_in 'user_password', :with => somebody.password
-      click_on 'Sign in'
+      sign_in_as(nil) # no role
```
kgf, line 7: I like the comment, it helps

spec/features/authorization_skill_spec.rb
```diff
@@ -1,38 +1,27 @@
...
-      click_on 'Sign in'
-    end
+    before (:each) { sign_in_as(nil) } # no role
+
     it "cannot view skills" do
       skill = create(:skill)
       visit people_path
       expect(page).not_to have_content('Edit') #Need to scope this, or it will fail on Edith
       expect(page).not_to have_content(skill.name)
       expect(page).to have_content("Access Denied")
```
kgf: You moved this in an earlier file, should we move it here too ?

spec/features/authorization_spec.rb
```diff
@@ -1,17 +1,12 @@
 require 'rails_helper'
 
-      #save_and_open_page
```
kgf, line 3: In case you were wondering, this are littered about so I can copy and paste when I needed them. There should go now, though.

kgf:
Few small things...

spec/features/authorization_editor_spec.rb
david135: In general, I think so, though in this case they are nested, since for both the `reader` tests and the `editor` tests we want a `person`, but we want to be signed in as a different role for each one.  Some documentation [here](http://www.relishapp.com/rspec/rspec-core/docs/hooks/before-and-after-hooks#%60before%60/%60after%60-context-blocks-are-run-once).

spec/features/authorization_skill_spec.rb
david135: No problem.

spec/features/authorization_spec.rb
david135: Makes sense! Thanks.

spec/features/authorization_editor_spec.rb
kgf: Sorry, missed the indentation

Issue #97, opened 2016-03-15
Add groups to Gemfile


kgf:
Groups should be added to exclude gems that are not needed. This would speed up our test suite on Travis

OrthoDex:
Hey @kgf , would it be alright if I issue a quick PR for this? It'll help me get a little more familiar with the project. Further, it'll help out other subsequent PRs.

Apart from adding groups, we could cache the dependencies as explained [here](https://docs.travis-ci.com/user/caching/), and run parallel tests if possible explained [here](https://docs.travis-ci.com/user/speeding-up-the-build/).

kgf:
That sounds great to me. I've read some on the parallel builds, not the dependencies. (yet)
Thanks

OrthoDex:
Hey @kgf , on further reading I found that caching and parallel tests are both currently available only for private projects :disappointed: 
Sorry about that.

OrthoDex:
Hey @kgf , issued a PR at #204 . The test suite is failing on a nil class error. Pretty weird, I haven't touched any code that should affect the existing features.

kgf:
I have started a re-run on the tests

kgf:
fixed by #204

PR #204, opened 2016-10-17
Added gemfile to groups. Enabled built in caching of travis ci

OrthoDex:
I'll look into speeding up the travis build further, maybe through [knapsack](https://github.com/ArturT/knapsack) or [parallel_tests](https://github.com/grosser/parallel_tests).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,13 @@ services:
 language: ruby
 rvm:
   - 2.3.1
+cache: bundler
 before_script:
   - cp config/database.yml.travis config/database.yml
   - psql -c 'create database travis_ci_test;' -U postgres
 # uncomment this line if your project needs to run something other than `rake`:
 script: bundle exec rake db:schema:load && bundle exec rspec spec
+bundler_args: --without development production doc
 branches:
   only:
     - development
--- a/Gemfile
+++ b/Gemfile
@@ -21,8 +21,6 @@ end
 gem 'devise', '~> 3.0'
 
 group :test, :development do
-  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
-  gem 'byebug'
   gem 'thin'
   gem "rspec-rails"
   # gem 'rspec-activemodel-mocks' # TODO: remove if we no longer need mock_model or stub_model
@@ -33,7 +31,6 @@ group :test, :development do
   gem "mailcatcher"
   gem "launchy"
   gem "letter_opener"
-  gem 'better_errors'
   gem 'binding_of_caller'
   gem 'meta_request'
   gem 'database_cleaner', '1.0.1'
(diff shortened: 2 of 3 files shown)


kgf:
I re-ran the tests on Travis, and they failed again, same error:

```
1) Channels visits a display page
     Failure/Error: <h2><m%= @channel.person.fullname %>  <%= @channel.name %></h2>

     ActionView::Template::Error:
       undefined method `fullname' for nil:NilClass
     # ./app/views/channels/show.html.erb:8:in `_app_views_channels_show_html_erb___2007759606707425590_71127740'
     # ./app/controllers/channels_controller.rb:21:in `show'
     # ./spec/support/authorization_view_helper.rb:31:in `block in get_basic_editor_views'
     # ./spec/support/database_cleaner.rb:10:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # NoMethodError:
     #   undefined method `fullname' for nil:NilClass
     #   ./app/views/channels/show.html.erb:8:in `_app_views_channels_show_html_erb___2007759606707425590_71127740'
```

Do these work for you locally, @OrthoDex ?

kgf:
I'm not seeing a relation to the PR....

OrthoDex:
Hey @kgf, so better_errors was causing the failure, I wasn't really expecting that :thinking:  I assumed it would be capistrano related. I put it back in it's grouping, and the tests are passing locally and on travis. I'll try and see why better_errors would be causing this issue, doesn't make sense to me atm.

kgf:
I saw, good work.
Should I hold on this PR then ?

OrthoDex:
Turns out this is an open [issue](https://github.com/charliesome/better_errors/issues/304) on better_errors 

I tried removing web_console but the tests still don't pass. I guess we'll have to wait for a fix before moving better_errors.

OrthoDex:
Alright, so doing this removes the failure:

> #config/environments/test.rb
> config.action_dispatch.show_exceptions = true

Although I feel doing this just to make the tests pass is unneeded. 

Thoughts @kgf ?

kgf:
Thanks for the link. My reading of that issue is that it caused tests to pass that should have failed. We have the opposite issue, tests that should pass are failing. Also, the comment from 

> ppdeassis commented on Jun 19, 2015
> suggested that adding `binding_of_caller` also corrected it for him ?

OrthoDex:
Hey @kgf , putting binding_of_caller along with better_errors in development still results in failure.
image

kgf:
Drat. So what are the options?

OrthoDex:
Well,
1. We keep better_errors and binding_of_caller as it is. I'll try to see if I can contribute to fixing the issue.
2. We do this: 
`#config/environments/test.rb
config.action_dispatch.show_exceptions = true`

According to the rails page:

> ActionDispatch::ShowExceptions rescues any exception returned by the application and renders nice exception pages if the request is local or if config.consider_all_requests_local is set to true. If config.action_dispatch.show_exceptions is set to false, exceptions will be raised regardless.

This should make better errors not respond to any exceptions in that particular test, since there are already rendered exception pages.

kgf:
Let's go with option 2 for now. Thanks for your help !

david135:
Am I right in thinking that 'better_errors' should not be in 'test', regardless of this problem?  [They](https://github.com/charliesome/better_errors) seem to recommend it for 'development' only.

kgf:
I would agree, there's no one there to see it !

OrthoDex:
Hey @kfg , so I have made the needed changes, looks good to go. :smile:

kgf:
Great, I'll take a look in a few minutes

kgf:
Thanks for all your diligence on this.

Kevin

kgf:
This fixes #97

PR #205, opened 2016-10-17
148 build out notifications

amandamunoz:


amandamunoz:
I noticed the names that were initialized here are different than those in the wiki - should I change to wiki to match?

amandamunoz:
Also, I added it to the nav but can comment it out until it's functional

kgf:
I'm mobile right now, can't review completely, but....
Make this match the wiki, the wiki is more correct.

Add it to the nav, but permission it to manager only. No end user has that role. Permissions are handled in the ability model.
Not in this branch, but we should add a dev role perhaps, for just this sort of thing.

Gemfile
```diff
@@ -17,6 +17,9 @@ source 'https://rails-assets.org' do
   gem 'rails-assets-datetimepicker'
 end
 
+# Handles text messages and voice calls
+gem 'twilio-ruby'
```
kgf: Please pin this to the current version, allowing point releases. I have had issues where gems have changed unexpectedly and caused issues.

app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,62 @@
+class NotificationsController < ApplicationController
+  before_action :set_notification, only: [:show, :edit, :update, :destroy]
+
+  # GET /notifications
+  def index
+    @notifications = Notification.all
+    @page_title = "All Notifications"
```
kgf, line 8: This is good. Could you please add `@page_title` for the other actions ? Could just set it to `subject`, perhaps in a before action.

app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,62 @@
+class NotificationsController < ApplicationController
+  before_action :set_notification, only: [:show, :edit, :update, :destroy]
```
kgf, line 3: You'll need to include the calls for our authorization (cancancan) Examples are here
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/controllers/certs_controller.rb#L2

app/services/notification_text_message.rb
```diff
@@ -0,0 +1,23 @@
+class NotificationTextMessage
```
kgf: I would like you to please remove this file from this PR. I apologize, since I caused this confusion. The class is not properly named, and is about `message`s, not `notification`s. 
We'll want this file to be added once we get to message model.

app/views/notifications/_form.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  <%= f.error_notification %>
+
+  <div class="form-inputs">
+    <%= f.association :event %>
+    <%= f.input :status, as: :select, collection: Notification::VALID_STATUSES %>
+    <%= f.input :hours_to_try %>
+    <%= f.input :minutes_interval %>
+    <%= f.input :attempts_before_escalation %>
+    <%= f.input :start_at %>
+    <%= f.input :started_at %>
```
kgf: This should not be on the form, it is filled in by the system.
It would be nice, but not required, if you could add some hints and/or labels to the inputs.

app/views/notifications/edit.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1>Editing Notification</h1>
+
+<%= render 'form' %>
+
+<%= link_to 'Show', @notification %> |
```
kgf: If you could please remove these links, as we use a sidebar for navigation, in this style
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/views/certs/edit.html.erb#L3

app/views/notifications/index.html.erb
```diff
@@ -0,0 +1,28 @@
...
+        <td><%= notification.channels %></td>
+        <td><%= link_to 'Edit', edit_notification_path(notification) %></td>
+      </tr>
+    <% end %>
+  </tbody>
+</table>
+
+<br>
+
+<%= link_to 'New Notification', new_notification_path %>
```
kgf: This should also be in the sidebar. Thanks

app/views/notifications/show.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  <strong>Start at:</strong>
+  <%= @notification.start_at %>
+</p>
+
+<p>
+  <strong>Channels:</strong>
+  <%= @notification.channels %>
+</p>
+
+<%= link_to 'Edit', edit_notification_path(@notification) %> |
```
kgf: Another to move, please.

db/migrate/20160927022134_create_notifications.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.belongs_to :event, index: true, foreign_key: true
+      t.string :status
+      t.string :subject
+      t.string :body
+      t.integer :hours_to_try
+      t.integer :minutes_interval
+      t.integer :attempts_before_escalation
+      t.datetime :start_at
+      t.datetime :started_at
+      t.string :channels
```
kgf: Since we'll probably want to store the channels as a serialized array, I think this should be text
http://stackoverflow.com/questions/6694432/using-rails-serialize-to-save-hash-to-database

db/migrate/20161009235659_edit_notifications.rb
```diff
@@ -0,0 +1,10 @@
+class EditNotifications < ActiveRecord::Migration
+  def change
+    remove_column :notifications, :subject, :string
+    remove_column :notifications, :body, :string
+    add_column :notifications, :groups, :string
```
kgf: Same as above, I suspect `groups` will be serialized.

spec/controllers/notifications_controller_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+# Message expectations are only used when there is no simpler way to specify
+# that an instance is receiving a specific message.
+
+RSpec.describe NotificationsController, type: :controller do
+
+  # This should return the minimal set of attributes required to create a valid
+  # Notification. As you add validations to Notification, be sure to
+  # adjust the attributes here as well.
+  let(:valid_attributes) {
+    skip("Add a hash of attributes valid for your model")
```
kgf: I'm not sure, but I think not setting these will result in these tests not working.

spec/factories/notifications.rb
```diff
@@ -0,0 +1,7 @@
+FactoryGirl.define do
+  factory :notification do
+    status "sent"
```
kgf: I don't think this factory is valid, since that's not one of the `VALID_STATUSES`

spec/factories/notifications.rb
```diff
@@ -0,0 +1,7 @@
...
+    started_at "2016-09-26 22:21:34"
+    channels "Email Text"
```
kgf, line 5: Perhaps use an array ?

spec/views/notifications/edit.html.erb_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe "notifications/edit", type: :view do
+  before(:each) do
+    @notification = assign(:notification, Notification.create!(
```
kgf, line 5: Do you not want to use the factory here ?

spec/views/notifications/index.html.erb_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe "notifications/index", type: :view do
+  before(:each) do
+    assign(:notifications, [
+      Notification.create!(
+        :event => nil,
+        :status => "active",
```
kgf, line 8: These statuses too

app/models/notification.rb
```diff
@@ -0,0 +1,9 @@
+class Notification < ActiveRecord::Base
+  belongs_to :event
+  # has_many :recipients
+
+  VALID_STATUSES = %w(pending active canceled complete expired)
```
kgf, line 5: Since these values are being used in the UI, please make them proper case.

app/views/notifications/_form.html.erb
```diff
@@ -0,0 +1,18 @@
+<%= simple_form_for(@notification) do |f| %>
+  <%= f.error_notification %>
+
+  <div class="form-inputs">
+    <%= f.association :event %>
+    <%= f.input :status, as: :select, collection: Notification::VALID_STATUSES %>
```
kgf, line 6: This is technically correct and I like how you accomplished it, but status should probably be managed by the controller. People can probably only choose **Pending** **Active** and **Canceled**. Then there is a bit of trickiness around changing status. A new Notification can only be changed to pending or active. A pending one can only move to active or canceled. You could capture the state machine in a hash, using the current state as the key, and the possible transitions as the values ?

app/views/notifications/index.html.erb
amandamunoz: I can add the sidebar here but i'm noticing it's not present for any other objects - should they have them as well?

kgf: Actually they all already do, as it's in the application layout.
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/views/layouts/application.html.erb#L18-L19
Sidebars for everybody !

Issue #155, opened 2016-09-30
Add Requirements to Tasks


kgf:
Requirements will need to be linked to a task. They will need the following attributes, at a minimum.

link to task
link to assignment, which doesn't exist yet
minimum_people, an integer
maximum_people, an integer
desired_people, an integer
skill, a reference to the skills table
optional, a boolean
floating, a boolean

kgf:
This is now present on development !

PR #207, opened 2016-10-20
requirements model v0.2

david135:
Addresses #155. This begins implementation of the Requirement model.  It will ultimately be part of the system linking Event->Task->Requirement->Assignment->Person.  Some of its functionality will need to adjust as the Assignment system is implemented.

### DIFF
--- a/app/assets/javascripts/requirements.coffee
+++ b/app/assets/javascripts/requirements.coffee
@@ -1,19 +1,13 @@
 jQuery ->
-  $('#tasks').dataTable
+  $('#requirements').dataTable
     columnDefs: [
                   { orderable: false, targets: [-1] },
                 ]
 
 jQuery ->
-  $('#event_tasks').dataTable
+  $('#task_requirements').dataTable
     ## https://datatables.net/examples/basic_init/scroll_y_dynamic.html
     scrollY: '20vh',
     columnDefs: [
                   { orderable: false, targets: [-1] },
                 ]
-
-  $('a[data-toggle="tab"]').on 'shown.bs.tab', (e) ->
-    $.fn.dataTable.tables(
-      visible: true
-      api: true).columns.adjust()
-    return
--- a/app/models/requirement.rb
+++ b/app/models/requirement.rb
@@ -1,0 +1,74 @@
+class Requirement < ActiveRecord::Base
+  belongs_to :task
+  belongs_to :skill
+  belongs_to :title
+  #has_many :assignments
+
+  validates :task, presence: true
+
+  # validates :skill, presence: title.blank?
+  # validates :title, presence: skill.blank?
+  validate  :valid_title_or_skill
+
+  validates :minimum_people, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validates :maximum_people, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validates :desired_people, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validate :valid_people_numbers
+
+  STATUS_CHOICES_ARRAY = ['Empty', 'Inadequate', 'Adequate', 'Satisfied', 'Full']
+  STATUS_CHOICES = STATUS_CHOICES_ARRAY.map.with_index { |v, i| [v, i] }.to_h
+
+  PRIORITIES = [['High', 1], ['Medium', 2], ['Low', 3]]
+
+  def priority_str
+    PRIORITIES[priority - 1][0] if priority.present?
+  end
+
+  def status
+    case
+    when number_filled >= maximum_people
+      return 'Full'
+    when number_filled >= desired_people
+      return 'Satisfied'
+    when number_filled >= minimum_people
+      return 'Adequate'
+    when number_filled > 0
+      return 'Inadequate'
+    else # number_filled == 0
+      return 'Empty'
+    end
+  end
+
+  def status_value
+    # return a value for sorting
+    return STATUS_CHOICES[status]
+  end
+
+  def number_filled
+    # temporary, until assignments are implemented
+    @stub_number_filled ||= rand(maximum_people + 1)
+    return @stub_number_filled
+  end
+
+  private
+    def valid_people_numbers
+      if (minimum_people.present? and maximum_people.present? and desired_people.present?)
+        if maximum_people < minimum_people
+          errors.add(:maximum_people, "must be greater-than or equal to 'minumum_people")
+        end
+        if desired_people < minimum_people
+          errors.add(:desired_people, "must be greater-than or equal to 'minumum_people")
+        end
+        if desired_people > maximum_people
+          errors.add(:desired_people, "must be less-than or equal to 'maxumum_people")
+        end
+      end
+    end
+
+    def valid_title_or_skill
+      if title.blank? && skill.blank?
+        errors.add(:title, "must have a title or a skill")
+        errors.add(:skill, "must have a title or a skill")
+      end
+    end
+end
--- a/app/models/skill.rb
+++ b/app/models/skill.rb
@@ -2,10 +2,11 @@ class Skill < ActiveRecord::Base
   attr_accessible :name, :status, :course_ids, :title_ids
   has_and_belongs_to_many :courses
   has_and_belongs_to_many :titles
+  has_many :requirements
   validates_presence_of :name, :status
 
   scope :active, -> { where( status: "Active" ) }
-  
+
   def to_s
     self.name
   end
--- a/app/views/requirements/edit.html.erb
+++ b/app/views/requirements/edit.html.erb
@@ -1,0 +1,9 @@
+<% content_for :sidebar do %>
+  <h2>Edit Requirement</h2>
+  <h3>(for <%= @requirement.task.title %>)</h3>
+  <%= sidebar_button_link 'Return to Task', @requirement.task %>
+  <%= sidebar_button_link 'Return to Requirements List', requirements_path %>
+  <%= sidebar_button_link 'Cancel Edit', :back %>
+<% end %>
+
+<%= render 'form' %>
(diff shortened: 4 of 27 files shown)


app/models/requirement.rb
```diff
@@ -0,0 +1,57 @@
+class Requirement < ActiveRecord::Base
+  belongs_to :task
+  belongs_to :skill
+  #has_many :assignments
```
kgf, line 5: Will need to add belongs_to title.

app/models/requirement.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  validates :task, presence: true
+  validates :skill, presence: true
```
kgf: This will get harder. You need at least one of title or skill.

app/models/requirement.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  validates :minimum_people, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validates :maximum_people, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validates :desired_people, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+  validate :valid_people_numbers
+
```
kgf, line 17: Nice validations !

app/models/requirement.rb
```diff
@@ -0,0 +1,57 @@
...
+  STATUS_CHOICES_ARRAY = ['Empty', 'Inadequate', 'Adequate', 'Satisfied', 'Full']
```
kgf, line 18: What is the value of `STATUS_CHOICES_ARRAY` ? It seems to only be used to build the hash `STATUS_CHOICES`.

app/models/skill.rb
```diff
@@ -2,6 +2,7 @@ class Skill < ActiveRecord::Base
   attr_accessible :name, :status, :course_ids, :title_ids
   has_and_belongs_to_many :courses
   has_and_belongs_to_many :titles
+  has_many :requirements
```
kgf, line 5: Similar line will need to be added to title.

app/views/requirements/edit.html.erb
```diff
@@ -0,0 +1,9 @@
+<% content_for :sidebar do %>
+  <h2>Edit Requirement</h2>
+  <h3>(for <%= @requirement.task.title %>)</h3>
+  <%= sidebar_button_link 'Return to Task', @requirement.task %>
+  <%= sidebar_button_link 'Return to Requirements List', requirements_path %>
+  <%= sidebar_button_link 'Show Requirement', @requirement %>
```
kgf: Please change this to `Cancel Edit`, which is what it does. It would be ideal if it returned from where they started.
I'll probably try to change this to `Cancel Edit` in all the places.

app/views/requirements/new.html.erb
```diff
@@ -0,0 +1,7 @@
+<% content_for :sidebar do %>
+  <h2>New Requirement</h2>
+  <%= sidebar_button_link 'Return to Task', @requirement.task %>
+  <%= sidebar_button_link 'Return to Requirements List', requirements_path %>
```
kgf: `Return to Event` probably makes more sense, rather than requirements list

app/views/requirements/show.html.erb
```diff
@@ -0,0 +1,35 @@
+<% content_for :sidebar do %>
+  <h2>Requirement</h2>
+  <%= sidebar_button_link 'Return to Task', @requirement.task %>
+  <%= sidebar_button_link 'Return to Requirements List', requirements_path %>
```
kgf: Same, return to event

app/views/requirements/show.html.erb
```diff
@@ -0,0 +1,35 @@
...
+  <span class="h3"><%= @requirement.skill %> for <%= link_to @requirement.task.title, task_path(@requirement.task) %></span>
+  (<%= link_to @requirement.task.event.title, event_path(@requirement.task.event) %> on <%= format_date_value(@requirement.task.start_time) %>)
+</p>
+
+Display option 2:
+<p>
+  <span class="h3"><%= @requirement.skill %></span> for
+  <span class="h4"><%= link_to @requirement.task.event.title, event_path(@requirement.task.event) %></span>:
+  <span class="h5"><%= link_to @requirement.task.title, task_path(@requirement.task)%></span>
+  <% if @requirement.task.event.location.present? %>
```
kgf: That's a lot of dots...
We should change this to be a method on task. The task location should be either a tasks's location, if defined, or the event's location.

config/routes.rb
```diff
@@ -35,7 +35,12 @@
   resources :courses
   resources :certs
 
-  resources :tasks, except: [:new, :create], constraints: { id: /\d+/ }
+  resources :requirements, except: [:new, :create], constraints: { id: /\d+/ }
```
kgf, line 39: That's nice routing.

db/migrate/20161013182014_create_requirements.rb
```diff
@@ -0,0 +1,15 @@
+class CreateRequirements < ActiveRecord::Migration
+  def change
+    create_table :requirements do |t|
+      t.references :task, index: true, foreign_key: true
+      t.references :skill, index: true, foreign_key: true
+      t.integer :minimum_people
```
kgf: Add title, as seen above

spec/factories/requirements.rb
```diff
@@ -0,0 +1,11 @@
+FactoryGirl.define do
+  factory :requirement do
+    association :task
```
kgf: I would prefer not to have the associations in the factory, but build them in the specs, as needed.

spec/models/requirement_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'rails_helper'
+
+RSpec.describe Requirement, type: :model do
+  let(:an_event) { build_stubbed(:event) }
+  let(:a_skill) { build_stubbed(:skill) }
+  let(:a_task) { build_stubbed(:task, event: an_event) }
```
kgf, line 5: Yes, this style.

app/models/requirement.rb
david135: I thought that maintaining the hash with the numerical values (used for sorting tables) might be burdensome (e.g. if I want to add a status in between or change the order, then I must change all the subsequent numbers). I also used this approach in the Task model.  Happy to hear suggestions for improved implementation here.

app/views/requirements/edit.html.erb
david135: I'll change this here, and can make a wider separate PR for other models.

app/models/requirement.rb
kgf: What is being stored in the db ? Cause if we're storing the index, changing the array would be really bad !

app/views/requirements/edit.html.erb
kgf: Sounds good, thanks

kgf:
Could you please also add `priority` to requirements ? Will be helpful in sorting, fill the most important things first.
Will also need it on tasks, but that should probably be a separate PR.

david135:
@kgf, do you think we need a separate `priority` for both Requirement and Task?

kgf:
Yes. Since the requirements are within a single task, I would want the most important requirements of the most important tasks.

app/models/requirement.rb
david135: See the `status` method (which relies on stubs for assignments).  We compare how many people we have assigned with the desired/required numbers and return the appropriate status.  As of now, we don't store any of this in the db. (If we think performance will be an issue, we can cache whatever we like, but then we'll have to take care when assignments are changed.)

app/assets/javascripts/requirements.coffee
```diff
@@ -0,0 +1,13 @@
+jQuery ->
+  $('#requirements').dataTable
+    columnDefs: [
+                  { orderable: false, targets: [-1] },
+                ]
+
+jQuery ->
+  $('#task_requirements').dataTable
+    ## https://datatables.net/examples/basic_init/scroll_y_dynamic.html
```
kgf, line 9: I appreciate these comments.

app/models/requirement.rb
```diff
@@ -0,0 +1,74 @@
+class Requirement < ActiveRecord::Base
+  belongs_to :task
+  belongs_to :skill
```
kgf, line 3: We may want `belongs_to :event, through task` so we can get back directly.

app/views/requirements/edit.html.erb
```diff
@@ -0,0 +1,9 @@
+<% content_for :sidebar do %>
+  <h2>Edit Requirement</h2>
+  <h3>(for <%= @requirement.task.title %>)</h3>
+  <%= sidebar_button_link 'Return to Task', @requirement.task %>
+  <%= sidebar_button_link 'Return to Requirements List', requirements_path %>
```
kgf, line 5: We can make a new issue out of this, but I don't think there's much value to requirements without the context of an event.

app/views/requirements/new.html.erb
```diff
@@ -0,0 +1,7 @@
+<% content_for :sidebar do %>
+  <h2>New Requirement</h2>
+  <%= sidebar_button_link 'Return to Task', @requirement.task %>
+  <%= sidebar_button_link 'Return to Event', @requirement.task.event %>
```
kgf, line 4: If we add that through call, we can shorten this one dot.

app/views/tasks/show.html.erb
```diff
@@ -24,17 +25,33 @@
...
+        <%= render(:partial => "availabilities/availabilities_table",
+                   :locals => { :availabilities => ( @task.event.responses),
+                                :partial_availabilities => ( @task.event.partial_responses) }
+                  ) || "No Availabilities yet !" %>
 
-  <%= render(partial: "people/people_table",
-             locals: { people: @task.event.unresponsive_people,
-               title: "No Response",
-               row_class: 'class="warning"'}) || "Everyone Responded !" %>
+        <%= render(partial: "people/people_table",
```
kgf, line 51: Not sure yet, but we may only want the unresponsive on the event page. Once you're in the task, you're thinking about scheduling.

PR #210, opened 2016-10-21
'Cancel Edit' links go back

david135:
give uniform behavior to `edit` pages, with a 'Cancel Edit' link that goes `:back`


app/views/titles/edit.html.erb
```diff
@@ -1,3 +1,6 @@
-<h1>Editing title</h1>
+<% content_for :sidebar do %>
```
kgf, line 1: I appreciate that you added this where it had been missing before. Thanks

Issue #98, opened 2016-03-18
Person form add zip code look up


VanAuken:
On '**New Person Form**', and edit '**Person**' form, please add feature that '**city**' updates according to '**zip code**' entry.

This will make data entry easier for user, as they won't need to type the city.  It will also produce an opportunity for the user to recognize if they have misread the applicants handwriting.

kgf:
Should this also update the state field ?

kgf:
Ideally this should reference an online data source via AJAX

kgf:
USPS does seem to have an API for this
https://www.usps.com/business/web-tools-apis/address-information-api.htm

OrthoDex:
Hey! I'd like to take this up if it's alright, we can use the Google Geocode API: https://developers.google.com/maps/documentation/geocoding/start?csw=1

kgf:
@OrthoDex That would be great ! Thanks. 

Please reference the API key using env vars, similar to this:

```
# create Twilio Client
    account_sid = ENV["TWILIO_ACCOUNT_SID"] # Your Account SID from www.twilio.com/console
    auth_token = ENV["TWILIO_AUTH_TOKEN"]   # Your Auth Token from www.twilio.com/console
```

Please also scope them to google, like GOOGLE_ACCOUNT_API_KEY
I hope you'll be able to get this to work when giving Google only the zipcode.

Thanks again for your interest.

OrthoDex:
Alright! I'll issue a pull request soon. :smile:

kgf:
I'm looking forward to it, and so is @VanAuken !

kgf:
I am mentioning this on all the claimed issues. We always appreciate people's help, but if you could have any PRs submitted by 1PM EST this Sunday, I'll have time to review and integrate them for a scheduled demo on Monday.

Thanks !
Kevin

kgf:
Fixed by #218

PR #218, opened 2016-10-22
Added city and state autofill based on zip code entry

OrthoDex:
Hey @kgf , I added the working ajax code to auto fill the city and state values.


kgf:
Thanks for your PR, that's great. 
Two small issues.  
One, there is a merge conflict. 
Two, you added `dotenv` to the Gemfile. Understandable, given my requirement of `ENV` variables. However, we have some people using that at my day job, and we've had lots o problems with it not working correctly. What if we switched to `secrets.yml` ?

If you use slack, I could send you an invite to our team.
Thanks for your help !
Kevin

OrthoDex:
Hey @kgf , so I was going to use the secrets.yml, but it states [here](http://www.collectiveidea.com/blog/archives/2013/12/18/the-marriage-of-figaro-and-rails/) that secrets.yml should be checked into .gitignore.

We could do either of the following: 
1. I can put the API key in secrets.yml but can we check secrets.yml into .gitignore?
2. We can use the [figaro](https://github.com/laserlemon/figaro) gem.

kgf:
I would prefer option 1, try to keep it simple.
Thanks!

OrthoDex:
@kgf , did the needful :smile:

kgf:
Were you going to add `secrets.yml` to `.gitignore` ? I realize we have one commited now, but we probably shouldn't...
We can remove using this technique.
http://stackoverflow.com/questions/1274057/how-to-make-git-forget-about-a-file-that-was-tracked-but-is-now-in-gitignore
Thoughts ?

OrthoDex:
Hey @kgf , yes I wanted to add secrets.yml to .gitignore. We could use the link you provided or refer to [this](https://help.github.com/articles/remove-sensitive-data/)

I have used bfg before, and it works pretty well.

On a side note, I think the conflict in Gemfile.lock is because of the later version of my bundler gem, and I have changed people.js.coffee to people.js.coffee.erb to embed the Google Maps API Key.

kgf:
If you could resolve the merge conflict, I will merge this. We can get the .gitignore update on another pass.

Thanks for your help.
Kevin

kgf:
I had to go look up what BFG was, which wasn't easy when a recent movie of the same name had an actress named Ruby !
Anyway, I don't think we currently have any secrets in the repo to be cleaned. Although secrets is in there, it does not contain any sensitive data.
Do you have a different opinion ?

Thanks for your help,
Kevin
Reference http://bwillis.github.io/2014/05/20/clean-secrets-git-repo/

kgf:
I thought on this some more on my commute, and I realized that we probably don't want it in secrets either, but instead inside the database as a Setting. This way end-users can update these and not need to restart the server, or need developer assistance. ("You're telling me I need to just edit the YAML file ? Sure....")

OrthoDex:
Hey @kgf , sorry about not linking to BFG :disappointed_relieved: 
I was about to add it to a yaml file. There's a great railscasts on it [here](http://railscasts.com/episodes/85-yaml-configuration-file)

kgf:
I'm going to merge this to get the great work you've done this far into the codebase. Long term I want to move the API key into the settings model so that folks who are just using do not need to be able to ssh into the server to change the value. It should be possible through the UI.
One last question. what does this do if the api key is missing ?

Thanks for your help
Kevin

OrthoDex:
If the API key is missing, the ajax functionality won't work, and it'll throw an [error](https://developers.google.com/maps/documentation/javascript/error-messages#missing-key-map-error) in the console.

I am not really sure about moving the API key into the settings model, majorly because if anyone is able to generate the Key, they should be able to set the production environment variable as well.

Thoughts?

kgf:
I wanted to get that merged, not have you rebase again. However, I intend to create a new issue to have it not error if the API key is missing, but to do nothing. Perhaps you would be interested ?

Could you provide a link or a few lines on how to generate a Google key ?

Thanks for all your help, and your opinions.
Kevin

kgf:
Fixes #98

Issue #212, opened 2016-10-22
Change events listing to show partial availability


kgf:
Looking at this recent screenshot, we see that one person is available, and two have not responded. No one has reported unavailable yet. There are three small numbers at the top, outlined in green red and yellow. They currently mean Available, Unavailable, and Unknown. This issue looks to change that. The desired display is to have four numbers. A green background still means available, _then_ a yellow background, which means they're available for _part_ of the event; partially available. Then a red background which, like now, means they are unavailable. Finally, a number with a grey background, showing the number of people who have not responded. This would also want the styling of the `No Responses` table to be no background, and the Available table to have styling showing yellow for the folks partially available. That should probably make a call to the `Event` class to determine partial availability. Also, the size of these number squares should be increased to make it easier to see on a mobile client. These numbers are very important,

![screenshot](https://cloud.githubusercontent.com/assets/8333278/19616743/595de572-97e9-11e6-94d0-81a9db048afb.png)

_New Improved Version:_
<img width="1190" alt="screen shot 2016-10-22 at 11 00 43 am" src="https://cloud.githubusercontent.com/assets/2431417/19620237/dfa399da-9846-11e6-81c3-2a59357f2110.png">

kgf:
The ideal PR would be ready by 1PM EST tomorrow, so I could include it in a demo on Monday.

kgf:
This is the bit in the class that will need a new method
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/models/event.rb#L34-L48

colinfike:
taking a look now

kgf:
Thanks for your help!

colinfike:
No problem, happy to help

kgf:
Closed by #220

PR #220, opened 2016-10-23
Feature/change event listing

colinfike:
…ing helpers to create the proper labels/styling

There were a few ways I was thinking about doing this but I ended up going this route. Tried to cut down on duplication where possible.

Updated the for_time_span scope to only return users whose availability is a superset of the events time range. The partially_available scope only returns users who have partial availability. 

screen shot 2016-10-22 at 8 06 07 pm
screen shot 2016-10-22 at 8 06 13 pm

### DIFF
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -5,18 +5,18 @@ module EventsHelper
   end
 
   def display_event_status(event)
-    available    = event.availabilities.count.to_s
-    partial      = event.partial_availabilities.count.to_s
-    unavailable  = event.unavailabilities.count.to_s
-    no_response  = event.unresponsive_people.count.to_s
+    available            = event.availabilities.count.to_s
+    partially_available  = event.partial_availabilities.count.to_s
+    unavailable          = event.unavailabilities.count.to_s
+    no_response          = event.unresponsive_people.count.to_s
     content_tag(:div) {
       capture do
         concat event_status_label(event)
         concat content_tag(:span, 'Personnel:', class: 'event-labels')
-        concat make_label(available, 'label label-success event-labels', tooltip: 'available')
-        concat make_label(partial, 'label label-warning event-labels', tooltip: 'available')
-        concat make_label(unavailable, 'label label-danger event-labels', tooltip: 'unavailable')
-        concat make_label(no_response, 'label label-default event-labels', tooltip: 'no response')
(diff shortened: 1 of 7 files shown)


kgf:
Taking a look. Thanks for your help !

app/helpers/events_helper.rb
```diff
@@ -6,18 +6,17 @@ def event_status_label(event)
...
-        concat " (Personnel: "
-        concat make_label(available, 'label label-success', tooltip: 'available')
-        concat "/"
-        concat make_label(unavailable, 'label label-danger', tooltip: 'unavailable')
-        concat "/"
-        concat make_label(no_response, 'label label-warning', tooltip: 'no response')
-        concat ")"
+        concat content_tag(:span, 'Personnel:', class: 'event-labels')
+        concat make_label(available, 'label label-success event-labels', tooltip: 'available')
+        concat make_label(partial, 'label label-warning event-labels', tooltip: 'available')
```
kgf: `Partially Available` ? Proper case may also make the others easier to read.

app/models/event.rb
```diff
@@ -32,9 +32,22 @@ def to_s
   end
 
   def unavailabilities
-    responses.unavailable
+    responses.unavailable + partial_responses.unavailable
```
kgf, line 35: Ah, nice catch. I had been thinking of partial availabilities only in the context of Available, not Unavailable

app/helpers/events_helper.rb
colinfike: Updated the tooltips and also updated the partial variable name just to make it a bit more consistent

app/views/availabilities/_availabilities_table.html.erb
```diff
@@ -11,9 +11,10 @@
     </tr>
   </thead>
   <tbody>
-    <% availabilities.each do |availability| %>
-      <tr <%== availability_status_class(availability) %>>
-        <td><%= availability_status_label(availability) %></td>
+      <!-- Full available -->
```
kgf, line 14: What does this mean ?

kgf:
Thanks for your help.

I think we'll end up with a refactoring in this area, as we figure out Availabilitiy and what not, but this does what I need for the demo ! Thanks.

Kevin

kgf:
Fixes #212

colinfike:
Yeah it looked like some of that could be ironed out once some specs were tightened up but I'm glad it'll help you out for the demo!

PR #221, opened 2016-10-23
add priority and location to Task model

david135:


app/models/task.rb
```diff
@@ -9,13 +9,17 @@ class Task < ActiveRecord::Base
...
   # has_many :assignments, through: :requirements
 
-  STATUS_CHOICES_ARRAY = ['Empty', 'Partially Filled', 'Full', 'Cancelled']
-
   # The following provides the equivalent of:
-  # STATUS_CHOICES = { 'Empty' => 0, 'Partially Filled' => 1, 'Full' => 2, 'Cancelled' => 3 }
-  # This is used to give a sortable numeric value based on the order of the above array.
+  # STATUS_CHOICES = { 'Empty' => 0, 'Inadequate' => 1, 'Adequate' => 2, 'Satisfied' => 3, 'Full' => 4, 'Cancelled' => 5 }
+  STATUS_CHOICES_ARRAY = ['Empty', 'Inadequate', 'Adequate', 'Satisfied', 'Full', 'Cancelled']
   STATUS_CHOICES = STATUS_CHOICES_ARRAY.map.with_index { |v, i| [v, i] }.to_h
```
kgf, line 15: Still not sure about this line. I'll merge it, but it seems unnecessarily complex. You could

```
STATUS_CHOICES = { 'Empty' => 0, 'Inadequate' => 1, 'Adequate' => 2, 'Satisfied' => 3, 'Full' => 4, 'Cancelled' => 5 }
STATUS_CHOICES_ARRAY=STATUS_CHOICES.keys
```

Wouldn't that work ?

app/models/task.rb
```diff
@@ -9,13 +9,17 @@ class Task < ActiveRecord::Base
...
 
+  PRIORITIES = [['High', 1], ['Medium', 2], ['Low', 3]]
+
+  def priority_str
```
kgf, line 19: What if we handles this like the statuses

```
PRIORITIES = {1 => 'High', 2 => 'Medium', 3 => 'Low'}
PRIORITES[1] # ==> "High"
PRIORITIES.values # For form choices
```

Would not need the priority_str method then.
What do you think ?

app/views/tasks/_form.html.erb
```diff
@@ -3,7 +3,9 @@
 
   <div class="form-inputs">
     <%= f.input :title %>
+    <%= f.input :priority, as: :select, collection: Task::PRIORITIES %>
```
kgf, line 6: If you liked my earlier idea, you can use PRIORITIES.invert here

app/views/tasks/_tasks_table.html.erb
```diff
@@ -14,9 +16,11 @@
     <% tasks.each do |task| %>
       <tr <%== task_status_class(task) %>>
         <td><%= link_to task.title, task %></td>
+        <td <%== sorting_order(task.priority) %>><%= task_priority_label(task) %></td>
         <td><%= truncate(task.description) %></td>
+        <td><%= truncate(task.location) %></td>
```
kgf, line 21: Nice, thinking of the truncate

app/views/tasks/index.html.erb
```diff
@@ -8,7 +8,9 @@
     <tr>
```
kgf, line 8: We should probably drop all this. Tasks don't have a lot of value outside the context of an event.

After Monday 😄

Issue #213, opened 2016-10-22
New Person form change text on form from 'Sms' to 'Text Message (SMS)'


VanAuken:
New Person form change text on form from 'Sms' to 'Text Message (SMS)'.
This asking for a change in the static text promt for the check box.  Some of our data entry people will not recognise the term SMS.
While you're at it, please change text prompt 'Icsid' to 'ICS Id'.

PR #222, opened 2016-10-23
Update SMS and ICSID labels on people form

RamyElkest:
New Person form change text on form from 'Sms' to 'Text Message (SMS)' and 'icsid' to 'ICS id' #213


kgf:
Thanks so much for your help !

Kevin

Issue #216, opened 2016-10-22
.ico file needs to be changed from LIMS to ReadyResponder


VanAuken:
.ico file needs to be changed from LIMS to ReadyResponder
LIMS was old name of project.  Tab icon still reflects old name.  Change of .ico file will update icon.

BhaveshSGupta:
I can take this one

PR #224, opened 2016-10-23
Updated favicon

BhaveshSGupta:
#216


kgf:
When I look at the attached file, it seems to be the same graphic. @BhaveshSGupta, did you commit the wrong file ?

BhaveshSGupta:
I guess I may have, let update it, sorry for inconvenience, btw do have any logo, from which i can make a favicon or should i use the paint one which i have made.

kgf:
Unless you have an interest in graphic design, this may not be a great issue to take on. Perhaps another could tempt you ? If you do still have an interest, I will not be able to discuss this much before Tuesday, but I was thinking of something around the two **RR**s of Ready Responder.
Thanks for your interest !
Kevin

BhaveshSGupta:
I have updated PR please review it now, and it would be great if you could add hacktoberfest label to it thanx!

PR #226, opened 2016-10-23
Fix - Filter locations based on department

sampatbadhe:
Hello @kgf 

This PR will enhancement for filter locations based on department - https://github.com/ReadyResponder/ReadyResponder/issues/86

**Before**
screen shot 2016-10-23 at 11 32 08 am

**After**

screen shot 2016-10-23 at 11 31 35 am

Please review and advice.

Thanks


kgf:
Thanks, that does work much better.

Issue #233, opened 2016-10-23
On item edit, look to department setting for items


kgf:
The department model now has a setting for `We manage Items`. The new/edit item form should only show departments for which that setting is true.

colinfike:
I'll take this real quick

EDIT: Done in https://github.com/ReadyResponder/ReadyResponder/pull/235

PR #235, opened 2016-10-23
Only show departments on the item form where manage_items is true

colinfike:
Takes care of https://github.com/ReadyResponder/ReadyResponder/issues/233


kgf:
Thanks for the quick pr, and your help. I like the class method. 
I'll accept this as is, but could you add a test?

colinfike:
Added a test to make sure that the proper options are in the select tag

kgf:
Thanks for your help !

Issue #240, opened 2016-10-23
When displaying availabilities on events, partial availability shows twice.


kgf:
As shown by the screen shot below, when someone is partially available, it shows that availability twice. One of the columns is a link, and when I hover over it, I can see that they point to the same database record.
The ideal PR would include a test that would expose the bug.
<img width="906" alt="screen shot 2016-10-23 at 6 26 28 pm" src="https://cloud.githubusercontent.com/assets/2431417/19629977/6400bbe2-994e-11e6-814c-bd39021deac4.png">

EvanTedesco:
Can you possibly include steps to reproduce?  Sorry I just forked the repo but need to know how to re create the condition in order to fix it.  When I create an event, then create a single partial availability I only see it once.  Is the bug that the user can potentially create duplicate partial availabilities?

kgf:
They can create a second availability, because the code is not that resilient yet.. but, regarding this issue:
Create an event, note the start and end times.
Create an availability with the same times. (It will do this by default through the UI. It will show as partial, and it will show twice. It seems to be tied to #239

EvanTedesco:
OK I got it sorted thanks

kgf:
Fixed in #245

PR #245, opened 2016-10-24
partially_available? returns false for full availabilities

### LINKED ISSUE #239: When entering an availability, exact same time shows as partially available.
An event has a start time and an end time. If you enter an availability with those exact times, you are listed as `partially available`, not `available`. 
We do not need seconds in these times. A later issue will try to normalize all times to quarter hours.
As you can see from the screen shot, the event is from 7 until 9. Abigal is available during those times, but is only shown as partially available.
The ideal PR would include a test that would expose the bug.

screen shot 2016-10-23 at 6 29 03 pm

EvanTedesco:
this closes #239 and #240

### DIFF
--- a/app/models/availability.rb
+++ b/app/models/availability.rb
@@ -18,7 +18,7 @@ class Availability < ActiveRecord::Base
   scope :unavailable, -> { where(status: "Unavailable")}
 
   def partially_available?(event)
-    return false if event.nil?
+    return false if event.nil? || (event.end_time == self.end_time && event.start_time == self.start_time)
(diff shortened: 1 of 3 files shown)


kgf:
Thanks for your help !
This does address #239, as it now shows available, but the same record still shows twice. See the two Jane Does below.
Would you like me to merge this for 239, or wait for another commit ?

screen shot 2016-10-23 at 9 30 53 pm

EvanTedesco:
This should fix the uniqueness issue.  Would you mind giving me the hacktoberfest flair as well on the PR?  Sorry I don't think I can add labels.

kgf:
Thanks for your help! I think there may be a re factoring in this area, but this gets the correct results. Thanks!

kgf:
Fixes #240

PR #251, opened 2016-10-24
Use absolute url for Slack. Fixes #250.

### LINKED ISSUE #250: Slack link in README.md is relative path not absolute
Due to this clicking on the link just 404's, should be an easy fix

kushalpandya:
I've fixed the incorrect path for Slack team on [README.md](https://github.com/ReadyResponder/ReadyResponder/blob/development/README.md).


kgf:
Both those URLs work for me, but this is certainly better.

Thanks for your help !

PR #256, opened 2016-10-24
Show application date on the applicants view

### LINKED ISSUE #255: On the applicants view, show the application date
Add a column after department for this date. This attribute is available on the Person model. You could use the duration column of the People/Active view as an example of adding a date to the datatables.

moritzsternemann:
Resolves #255

### DIFF
--- a/app/views/people/applicants.html.erb
+++ b/app/views/people/applicants.html.erb
@@ -4,6 +4,7 @@
     <tr>
       <th>Name</th>
       <th>Department</th>
+      <th>Application date</th>
       <th></th>
     </tr>
   </thead>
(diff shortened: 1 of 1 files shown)


kgf (changes requested):
That is kinda a lot of logic in a view. Happily, we can reduce a lot of it due to business rules. We don't need to consider an application date in the future, so the elsif line can go. And line 21 just becomes static text then. And with no if, 23 goes.
Finally, this does not seem to add the hover text. You can see an example of that in the active view.

Thanks for your help.
Kevin

moritzsternemann:
Okay, didn't know about that.. 
With 'hover text' you mean the HTML title attribute on the table cell, which shows the plain date on hover, right?

kgf:
Yes, probably like this
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/views/people/index.html.erb#L30
Thanks for your help.

moritzsternemann:
That's actually in the original commit

+Removed the unnecessary code

kgf:
Doh! I see it now, right there on line 16. My mistake, sorry about that...
Regarding moving the extra lines. I see the new commit, but it seems to still have the lines for having a date in the future. In this use case, this shouldn't happen.
https://github.com/ReadyResponder/ReadyResponder/pull/256/commits/1a6409047316d54a31c0380fc088e8707f090c7d#diff-4cf1a79d5bd8df9c0db1c260fb53a830R22
It's quite close. Thanks for your help.
Kevin

kgf:
Thanks very much for your help.

PR #257, opened 2016-10-25
Make end_time's datepicker default to displaying the start_time

### LINKED ISSUE #252: After entering a start time on event, the end time should jump to a time at least as forward
Currently, if you choose a start time in a later month, you have to scroll to that later month again for the end date. Most events will start and end on the same day, so we might as well jump there. 

If the end date is already later than the start, do nothing.

jeffhos:
Every time the start date is changed on an Event, update the end time datepicker's startDate, so the datepicker will default to displaying a reasonable initial date.  Fixes #252

### DIFF
--- a/app/assets/javascripts/events.js.coffee
+++ b/app/assets/javascripts/events.js.coffee
@@ -17,3 +17,9 @@ jQuery ->
       visible: true
       api: true).columns.adjust()
     return
+
+  $('#event_start_time').change ->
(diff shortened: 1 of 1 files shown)


kgf:
I will review this in the morning. Thanks for your help.

kgf:
Thanks a lot for your help. Javascript is not my strong skill.

Kevin

Issue #10, opened 2013-09-13
Implement paper_trail for audit trail


kgf:
This does not need to happen just yet, but very soon.

https://github.com/airblade/paper_trail

kgf:
OK, now would be a good time !
This should be on the person, item, event models at a minimum.

Spandana-Sai:
Work in progress @kgf

kgf:
You're amazing, thanks for all your efforts.

PR #258, opened 2016-10-25
Implemented paper trail for models

Spandana-Sai:
- Item, Person, Event
#10

### DIFF
--- a/config/initializers/paper_trail.rb
+++ b/config/initializers/paper_trail.rb
@@ -1,0 +1,2 @@
+PaperTrail.config.track_associations = false
+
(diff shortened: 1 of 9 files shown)


config/initializers/paper_trail.rb
```diff
@@ -0,0 +1 @@
+PaperTrail.config.track_associations = false
```
kgf: Please add the new line char. I have actually had issues when it was missing!
Thanks

kgf:
One small change. Thanks for your help.

config/initializers/paper_trail.rb
Spandana-Sai: @kgf  done

kgf:
That closes an old one. 
Thanks for your help.

PR #265, opened 2016-10-28
adds TimeRoundingService to adjust times, fixes #259

### LINKED ISSUE #259: Create a service object to adjust times.
Given a start time and an end time, it should adjust them.
It should first remove the seconds and microseconds so we don't move times falsely.
It should adjust the start time to the nearest quarter hour after the the time given.
It should adjust the end time to the nearest quarter hour before the the time given.
If either time is already on the quarter hour, it should do nothing.
All times returned should not have a seconds or micro-seconds component, minute is specific enough.
This service should accept an object, such as event or availability. It should not care what type of object, only that it has those attributes.

_Please_ include tests.

Thanks for your help !

abrophy:
First time using RSpec, let me know if the tests need work, or if you have input on style used in services object.

Also if you would rather name the service object something else.

### DIFF
--- a/services/time_rounding_service.rb
+++ b/services/time_rounding_service.rb
@@ -1,0 +1,34 @@
+class TimeRoundingService
+
+  def initialize(object_to_adjust)
+    @adjustee = object_to_adjust
+  end
+
+  def closest_fifteen
+    truncate_seconds
+    @adjustee.start_time = nearest_fifteen_after(@adjustee.start_time)
+    @adjustee.end_time = nearest_fifteen_before(@adjustee.end_time)
+    @adjustee
+  end
+
+  private
+
+  def truncate_seconds
+    @adjustee.start_time = @adjustee.start_time.change(:usec => 0, :sec => 0)
+    @adjustee.end_time = @adjustee.end_time.change(:usec => 0, :sec => 0)
+  end
+
+  def nearest_fifteen_after time
+    minutes = time.min
+    minutes = ((minutes / 60.0 * 4).ceil / 4.0 * 60).to_i
+    time.change(:min => minutes)
+  end
+
+  def nearest_fifteen_before time
+    minutes = time.min
+    minutes = ((minutes / 60.0 * 4).floor / 4.0 * 60).to_i
+    time.change(:min => minutes)
+  end
+
+end
+
--- a/spec/services/time_rounding_service_spec.rb
+++ b/spec/services/time_rounding_service_spec.rb
@@ -5,57 +5,20 @@ RSpec.describe TimeRoundingService do
   describe "#closest_fifteen" do
 
     context "given an event" do
-      it "removes milliseconds from start_time" do
-        @event = build(:event)
+
+      it "rounds times to the closest quarter hour" do
+        @event = build_stubbed(:event)
+        @event.start_time = @event.start_time.change(:hour => 9, :min => 1,
+                                                     :sec => 35, :usec => 231)
+        @event.end_time = @event.end_time.change(:hour => 9, :min => 14,
+                                                     :sec => 35, :usec => 231)
+
         @rounded_event = TimeRoundingService.new(@event).closest_fifteen
-        expect(@rounded_event.start_time.strftime("%L")).to eq("000")
+
+        expect(@rounded_event.start_time.strftime("%H:%M:%S:%L")).to eq("09:15:00:000")
+        expect(@rounded_event.end_time.strftime("%H:%M:%S:%L")).to eq("09:00:00:000")
       end
 
-      it "removes seconds from start_time" do
-        @event = build(:event)
-        @rounded_event = TimeRoundingService.new(@event).closest_fifteen
-        expect(@rounded_event.start_time.strftime("%S")).to eq("00")
-      end
-
-      it "removes milliseconds from end_time" do
-        @event = build(:event)
-        @rounded_event = TimeRoundingService.new(@event).closest_fifteen
-        expect(@rounded_event.end_time.strftime("%L")).to eq("000")
-      end
-
-      it "removes seconds from end_time" do
-        @event = build(:event)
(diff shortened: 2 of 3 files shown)


services/time_rounding_service.rb
```diff
@@ -0,0 +1,34 @@
+class TimeRoundingService
+
+  def initialize(object_to_adjust)
+    @adjustee = object_to_adjust
+  end
+
+  def closest_fifteen
```
kgf, line 7: Traditionally, service objects only have a call method. I actually am coming to prefer breaking with tradition, as this does a better job of revealing the intention when you call it.

services/time_rounding_service.rb
```diff
@@ -0,0 +1,34 @@
...
+    truncate_seconds
+    @adjustee.start_time = nearest_fifteen_after(@adjustee.start_time)
+    @adjustee.end_time = nearest_fifteen_before(@adjustee.end_time)
+    @adjustee
+  end
+
+  private
+
+  def truncate_seconds
+    @adjustee.start_time = @adjustee.start_time.change(:usec => 0, :sec => 0)
```
kgf, line 17: Didn't know about `Time.change`. Nice

spec/services/time_rounding_service_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+
+RSpec.describe TimeRoundingService do
+
+  describe "#closest_fifteen" do
+
+    context "given an event" do
+      it "removes milliseconds from start_time" do
+        @event = build(:event)
+        @rounded_event = TimeRoundingService.new(@event).closest_fifteen
+        expect(@rounded_event.start_time.strftime("%L")).to eq("000")
```
kgf: Others would debate it, but I would be OK with the assertion for start_time to be with the assertion for end_time. Less set-up, fast test suite.

spec/services/time_rounding_service_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+      end
+
+      it "removes seconds from start_time" do
+        @event = build(:event)
+        @rounded_event = TimeRoundingService.new(@event).closest_fifteen
+        expect(@rounded_event.start_time.strftime("%S")).to eq("00")
+      end
+
+      it "removes milliseconds from end_time" do
+        @event = build(:event)
```
kgf: This line is common to many of your tests. You may want to take a look at `let`
https://www.relishapp.com/rspec/rspec-core/docs/helper-methods/let-and-let

spec/services/time_rounding_service_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+
+      it "rounds end time to closest quarter hour before given time" do
+        @event = build(:event)
+        @event.end_time = @event.end_time.change(:min => 2)
+        @rounded_event = TimeRoundingService.new(@event).closest_fifteen
+        expect(@rounded_event.end_time.strftime("%M")).to eq("00")
+      end
+    end
+
+    context "given an availability" do
```
kgf: I don't think it adds value to repeat the tests for a second class. There is no reference to the class in your code. 
Keeping the test suite fast is also a goal. In fact, I would consider fewer tests. Rather than testing each element of the times, go for equality. For example, create a start_time at, say, 9:10:32.760, and see if return value == 09:00:00.000. That tests all elements in one assertion, and is actually whats important. I will use this service later to know if one time equals another.

Your thoughts ?
Kevin

kgf:
Nice work, especially the provided tests ! Thanks

abrophy:
Thanks for the review, I'll make more commits to clean up these tests according to your suggestions

kgf:
Do you agree with my comments, esp the last one ?

david135:
While you're at it, I'd consider whether to use build or build_stubbed in new tests.  See [here](https://robots.thoughtbot.com/use-factory-girls-build-stubbed-for-a-faster-test) for some discussion.

abrophy:
Yes I do. I was actually concerned about whether I was adding too much overhead with how the tests are currently constructed, you gave a good alternative. I figured it would be easier to just drop the unnecessary tests if/when you brought this up.

As for tradition of service objects being for them just to be called, can you link me to an example for future reference?

abrophy:
build_stubbed looks attractive, I'm not sure how much of a speedup it will be though, If you find it suitable you should put up an issue asking someone to go through the test suite and update it where necessary.

david135:
I'm not sure this will DTRT with, e.g. 16:55.  Am I missing something?

kgf:
It depends on if 16:55 is a start time or an end time. If a start, it should go to 16:45. If an end, 17:00.

Does that help ?

abrophy:
Major speed up to tests but has the same coverage as before, hope you agree

services/time_rounding_service.rb
```diff
@@ -0,0 +1,34 @@
...
+    @adjustee.end_time = @adjustee.end_time.change(:usec => 0, :sec => 0)
+  end
+
+  def nearest_fifteen_after time
+    minutes = time.min
+    minutes = ((minutes / 60.0 * 4).ceil / 4.0 * 60).to_i
```
kgf, line 23: If you have 16:55, won't this return `60` for the minutes ? 
Could you add a test for that as well.

Probably the correct level of testing would be 4 contexts, one for each quarter of the hour, each with an assertion for the start time and an assertion for the end time.

Thanks
Kevin

abrophy:
oh man good catch @david135, on it

kgf:
I edited my comment regarding tests above

kgf:
Fascinating, this is listed as merged, as it contains the same code as @abrophy PR for messages. I did not explicitly merge this !

abrophy:
That's really strange, do you think it might have something to do with how I made my branches on the repo?

kgf:
I think it is just Github being smart. By merging the other, I effectively merged this.

PR #266, opened 2016-10-28
created the recipient model, links to person, fixes #165

### LINKED ISSUE #165: Add Recipients to Person model
This should follow the description in the wiki.

abrophy:


spec/models/recipient_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Recipient, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
kgf: This file is not helpful 😉

abrophy: on it

abrophy:
Let me know if there are other tests/validations you would like to add for recipient model

spec/factories/recipients.rb
```diff
@@ -0,0 +1,9 @@
+FactoryGirl.define do
+  factory :recipient do
+    association :notification, factory: :notification
```
kgf, line 3: I would prefer not to have the associations in the Factories. You can always create them as needed in the specs. However, when they're in the factory, you loose that control. 
Let's say you want to check that a validation on recipient is good; perhaps response_time can't be in the future, say. So you have the factory create one for you. Because of the associations, it will also create a notification and a person. And hey, notification factory lists an association to an Event ! Now you've created four objects to test one validation. 

That make sense ?
Thanks for your help,
Kevin

abrophy: You make a good point, I hadn't thought of the association factories creating complex hierarchies like that during testing.

I was using other factories in the project as an example, and several call on other factories for associations. How does one differentiate?

Would you avoid creating association with factory if:
- The related object isn't required for validation
- The related object has associations it calls on factories for

PR #267, opened 2016-10-28
adds the message model, fixes #198

### LINKED ISSUE #198: Create messages model, link to person.
See https://github.com/ReadyResponder/ReadyResponder/wiki/Notifications

The ideal PR would include tests for this model.

abrophy:


spec/factories/messages.rb
```diff
@@ -0,0 +1,5 @@
+FactoryGirl.define do
+  factory :message do
+
```
kgf: crickets

app/models/message.rb
```diff
@@ -0,0 +1,3 @@
+class Message < ActiveRecord::Base
+  belongs_to :recipient
```
kgf, line 2: belongs to channel ?

spec/models/message_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Message, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
kgf, line 4: Same as other PR

app/models/message.rb
abrophy: According to the messages entry in the wiki I thought that channel would be a string identifying what channel(s) the message was sent on. I see now that there's a model for channels, I apologize for my oversight, I'll correct this.

spec/factories/messages.rb
abrophy: Will add a commit with a factory for messages, I just wanted to get the model out as soon as possible as other issues rely on it.

kgf: Thanks !

app/models/message.rb
kgf: No problem. I've been trying to write better issues.

kgf:
Sadly, we still have a merge conflict

abrophy:
There we go,

Thank you for your patience, this is the first time I'm working in a distributed project like this and dealing with these issues.

spec/factories/messages.rb
```diff
@@ -0,0 +1,10 @@
+FactoryGirl.define do
+  factory :message do
+    subject "test subject"
+    status "Sent"
+    body "This is a test message!"
+    sent_at 2.hours.ago
+    association :recipient, factory: :recipient
```
kgf, line 7: I prefer to not have the associations in the factories. I find this slows the tests to much. There are many tests for messages that will not need the recipient or channel, and those that do can handle them. I have found that when the associations _are_ in the factory, it cascades. You asked for a recipient and a channel; but what if their factories have a person association ? And what if Recipient has an association to Notification ? You get the idea, you end up creating 12 objects for one test. And additional problem is that if recipient and channel created a Person, it would not be the same instance.
I am going to merge this to move forward, and create a new issue.

spec/services/time_rounding_service_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe TimeRoundingService do
+
+  describe "#closest_fifteen" do
+
+    context "given an event" do
+
+      it "rounds times to the closest quarter hour" do
```
kgf, line 9: This is good because both times are in the first quarter hour. Now to add three more for the other quarters, especially the last quarter. I will add a new issue for this.

services/time_rounding_service.rb
```diff
@@ -0,0 +1,34 @@
+class TimeRoundingService
```
kgf, line 1: This code does not seem to be related to `Adds the message model`, which is the title of this PR. Seems more like it snuck into a commit. Sadly, I beleive it still has the bug regarding the last quarter hour.
I suspect I should close the other PR, as this code will now be in the codebase ?

spec/services/time_rounding_service_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+        @event = build_stubbed(:event)
+        @event.start_time = @event.start_time.change(:hour => 9, :min => 1,
+                                                     :sec => 35, :usec => 231)
+        @event.end_time = @event.end_time.change(:hour => 9, :min => 14,
+                                                     :sec => 35, :usec => 231)
+
+        @rounded_event = TimeRoundingService.new(@event).closest_fifteen
+
+        expect(@rounded_event.start_time.strftime("%H:%M:%S:%L")).to eq("09:15:00:000")
```
kgf, line 18: Another way to do the assertion is to create a Time object, and test against that, rather than converting to strings.

kgf:
@abrophy No problem, glad we could help. Feel free with the questions.

Issue #20, opened 2014-08-23
Implement temperature heat map for attendance


kgf:
This would use something like Github's heatmap. There would be different color markers for when someone had worked, when someone responded that they were available, et al.

This is a great visualization, but it is a nice to have, thus a low priority.

kgf:
This could involve this component, for example.
https://developers.google.com/chart/interactive/docs/gallery/calendar

Spandana-Sai:
checking issue

kgf:
Thanks for all your help.

Spandana-Sai:
My Pleasure @kgf

kgf:
Sadly, while this did have the map, it does not function properly.

kgf:
Be the one to closer one of the oldest issues!

PR #268, opened 2016-10-28
Implement basic chart for attendance

Spandana-Sai:
- This is just an basic chart. Post discussion with @kgf. Will work further. #20

### DIFF
--- a/app/assets/javascripts/analytics.coffee
+++ b/app/assets/javascripts/analytics.coffee
@@ -1,0 +1,38 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
+
+drawChart = ->
+  dataTable = new google.visualization.DataTable()
+  dataTable.addColumn
+    type: 'date'
+    id: 'Date'
+  dataTable.addColumn
+    type: 'number'
+    id: 'Count'
+  $.ajax
+    url: '/analytics/calendar_chart'
+    type: 'GET'
+    dataType: 'json'
+    success: (data) ->
+      data = data[0]
+      for i of data
+        dataTable.addRow [
+          new Date(data[i].Date)
+          data[i].Count
+        ]
+      chart = new (google.visualization.Calendar)($('#chartContainer')[0])
+      options =
+        title: 'Availability Chart'
+        height: 350
+      chart.draw dataTable, options
+      return
+
+loadGoogle = ->
+  # define callback in load statement
+  google.charts.load 'current',
+    'packages': [ 'calendar' ]
+    'callback': drawChart
+
+if window.location.href.indexOf('calendar_chart') > -1
+  loadGoogle()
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/loader.js
```diff
@@ -0,0 +1,149 @@
+(function (){ var aa="\n//# sourceURL=",ba='<script type="text/javascript" src="',b="American Samoa",ca="Antigua and Barbuda",k="Bolivia",p="Bosna i Hercegovina",q="Botswana",da="British Virgin Islands",ea="Cayman Islands",fa="Christmas Island",ga="Falkland Islands",r="Ghana",ha="Guin\u00e9e \u00e9quatoriale",ia="Guyane fran\u00e7aise",t="Honduras",u="Indonesia",v="Itoophiyaa",ja="Kalaallit Nunaat",w="Kiribati",ka="LocaleNameConstants",la="Luxembourg",ma="Madagascar",na="Marshall Islands",x="Micronesia",oa="Moldova, Republica",
```
kgf: Where does this file come from ?

kgf:
Perhaps this could be on the person#show, giving what they had been available for in the last year ?

kgf:
@Spandana-Sai Github seems to have munged my comment.
I was trying to ask where `app/assets/javascripts/loader.js` comes from. It seems already minified.

app/assets/javascripts/loader.js
Spandana-Sai: @kgf This is nothing but     <script type="text/javascript" src="https://www.gstatic.com/charts/loader.js"></script>. Instead of script tag in application.html.erb. I have the file in  this path app/assets/javascripts/loader.js

Spandana-Sai:
@kgf . I thought there will be more maps for various stats that is the reason behind creating new controller. If we have only attendance map alone then I will change the logic to PeopleController

kgf:
I think we would really want both approaches. One is about how this person is doing, the other how the department as a whole is doing. 
Where did the loader file come from? Is that something Google provided?
Thanks for your help.

Spandana-Sai:
@kgf the file is provided by google.

kgf:
Howdy, thanks for your patience. I would like to use a CDN for the loader, rather than pushing it into the repo. I see from the sample code on [this site](https://developers.google.com/chart/interactive/docs/gallery/calendar#overview) that Google recommends `<script type="text/javascript" src="https://www.gstatic.com/charts/loader.js"></script>`, which will probably be a lot faster.
Thanks for your help.
Kevin

Spandana-Sai:
@kgf what I thought was whenever server was down gstatic.com then we can't load the map so that is the reason I have downloaded and added file in app/assets/javascritps. I will change the code.

kgf:
I'm going to merge this as it moves us in the right direction, but we'll need some more work on the controller action that loads the data. It causes a lot of DB queries, and I am suspect of the results.

Thanks for getting us this far !
Kevin

Spandana-Sai:
@kgf please let me know exactly what you need furthur and create an issue, assign it to me. I will work on that. Thank you :).

Issue #263, opened 2016-10-27
Refactor partial availability


kgf:
This does currently work, but I suspect it is not optimal. There is currently a bug in the four small status boxes showing the incorrect value.
The ideal PR would include tests.
If an Availability start time is less than or equal to the start time of the query, and the Availability end time is equal to or greater than the end time of the query, then it is "Available", Otherwise, I think it is unavailable.
I think this should be turned into a service object. It will be needed for both tasks and events, and perhaps others. It could be passed a Person (to get all their availabilities) and another object ( perhaps a task or an event, as long as it responds to start_time and end_time )
The code currently involved is around here => 
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/models/event.rb#L36

This one's for you @swaathi !
Thanks for your help.
Kevin

swaathi:
Thanks Kevin. Looking into this now.

kgf:
Great, thanks for your help.

swaathi:
When you said this,

There is currently a bug in the four small status boxes showing the incorrect value.

Do you know the scenario in which this happens? Does the seed file contain that data which causes this bug?

kgf:
Sorry, seeds.rb does not cause the bug. It could be related to the fact that we currently allow duplicate availabilities for the same time span. You can see the issue in the attached screen shot, which shows the entire availabilities table. 
<img width="897" alt="screen shot 2016-10-28 at 9 23 14 am" src="https://cloud.githubusercontent.com/assets/2431417/19807989/9edac640-9cf0-11e6-8e88-2d34bfb13a23.png">
Thanks for your help
Kevin

swaathi:
Awesome. This helps. Will try to fix it by Sunday. :)

_
Swaathi K
CTO, Skcript

kgf:
Thanks for your help !

swaathi:
No problem! I hope this addresses your bug. Feedback is welcome! :)

kgf:
Closed, will need a better description to make it actionable.

PR #277, opened 2016-10-30
Refactoring availability and adding tests - Fixing #263

swaathi:


swaathi:
Fixing #263.

app/models/availability.rb
```diff
@@ -21,4 +23,16 @@ def partially_available?(event)
...
   end
+
+  private
+  # Does not allow people to fill in times enclosing other times already claimed
+  # For Example,
+  # If Person A is available from 9:00 AM to 10:00 AM,
+  # He/She may not fill in another time from 9:15 to 10:15 AM
+  # However he/she may fill in from 10:00 AM to 10:15 AM
+  def must_not_enclose_previously_claimed_times
+    if person.availabilities.where("? > end_time AND end_time > ? OR ? > start_time AND start_time > ?", end_time, start_time, end_time, start_time).present?
```
kgf: Two things: 
Could you please break this line so it fits on Github ?
Please exclude Availabilities that are `cancelled`, since they don't count.

kgf:
Thanks for your work.
Sadly, the test suite does not pass. I took a very quick look, and the failures seemed in the Availabilities area.
Kevin

swaathi:
Hey Kevin. Yup, I'm actually looking into it. Unfortunately something messed up with my `postgresql` installation. I'm unable to replicate it in my dev machine because of that. Do you have any good resources to install it? I'm running on MacOS Sierra and used brew.

kgf:
That is how I installed it as well. Sorry, not a brew master
Kevin

app/models/availability.rb
swaathi: Sure thing. :)

swaathi:
Oh man. :( Thanks though!

spec/models/availability_spec.rb
```diff
@@ -34,4 +36,31 @@
       expect(availability.partially_available?(event)).to eq(false)
     end
   end
+
+  before(:each) do
+    availability = build(:availability, person: a_person, start_time: start_time, end_time: end_time)
+    availability.save
```
kgf, line 42: If you did create in line 41, would you need the save ? 
Also see @david135 comments on build_stubbed.

swaathi: I need to persist it to the database since the validations are written that it compares with what is stored. So if I store two availabilities with the same times, only the second throws the validation error.

kgf:
Sadly, this still does not pass tests, and it's been a while. I will happily re-open this if @swaathi has renewed interest.

PR #278, opened 2016-10-30
TimeRoundingService object updates

### LINKED ISSUE #274: TimeRounding service does not handle the last quarter hour correctly
This line has an issue when the minutes are between 45 and 59.
The PR addressing this should include tests for this code.
The correct level of testing would be 4 contexts, one for each quarter of the hour, each with an assertion for the start time and an assertion for the end time.

Also see #265 

https://github.com/ReadyResponder/ReadyResponder/blob/development/services/time_rounding_service.rb#L23

abrophy:
fixes #274


kgf:
Thanks for your help !

PR #280, opened 2016-10-30
Create commentable concern and link person to comments model

abhishekbose87:
Hi @kgf, I am facing a issue wherein the specs are failing because I already see a text column `comments` and that is clashing with the `Comment` model I am introducing.
Is there a use for keeping the text field ?

Also can you confirm that The new comments model has the following db design. Hope I got that part of the requirements ready.

```
  create_table "comments", force: :cascade do |t|
    t.string   "title",            limit: 50, default: ""
    t.text     "description"
    t.integer  "commentable_id"
    t.string   "commentable_type"
    t.integer  "person_id"
    t.datetime "created_at",                               null: false
    t.datetime "updated_at",                               null: false
  end
```

_TODO_: Display comments made by person on the person view page.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -1,5 +1,7 @@
 class Person < ActiveRecord::Base
   has_paper_trail
+  include Commentable
+
   include ActiveModel::ForbiddenAttributesProtection
   mount_uploader :portrait, PortraitUploader
 
--- a/app/views/people/_form.html.erb
+++ b/app/views/people/_form.html.erb
@@ -58,6 +58,6 @@
     <h4> Personnel Image </h4>
     <%= @person.portrait %>
     <%= f.file_field :portrait %>
-    <%= f.input :comments, :input_html => { :size => 40} %>
+    <%= f.input :old_comments, :input_html => { :size => 40} %>
     <%= f.button :submit %>
 <% end %>
--- a/spec/factories/comments.rb
+++ b/spec/factories/comments.rb
@@ -1,0 +1,5 @@
+FactoryGirl.define do
+  factory :comment do
+    
+  end
+end
--- a/spec/models/comment_spec.rb
+++ b/spec/models/comment_spec.rb
@@ -1,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Comment, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -164,7 +164,7 @@ class PeopleController < ApplicationController
       :city, :state, :zipcode, :start_date, :end_date,
       :application_date, :title, :gender, :portrait, :date_of_birth,
       :division1, :division2, :channels_attributes, :title_ids,
-      :title_order, :comments,
+      :title_order, :old_comments,
       :channels_attributes => [],
       :title_ids => [],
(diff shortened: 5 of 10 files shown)


kgf:
Thanks for your help! 
Sorry, this was more complex than I had first imagined. What you can do is write a migration to move the existing comments to something like old _ comments. Ideally the migration would move the old into the new format, however, renaming the column and changing the form should clear your path.
On the schema in your comment, I don't think you need person _id column. That would be contained in the commendable _id.
Hope this helps, feel free to ask more.
Thanks,
Kevin

app/models/person.rb
```diff
@@ -21,6 +22,7 @@ class Person < ActiveRecord::Base
   has_many :items, inverse_of: :owner, foreign_key: :owner_id
   has_many :inspections
   has_many :activities, as: :loggable
+  has_many :comments
```
kgf: If you use the concern, won't that already have this ?

spec/models/comment_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Comment, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
kgf, line 5: Sounds like good advice

kgf:
Perhaps you should get this working as polymorphic, then as a concern? I'll happily accept two PRs. The scope of the issue as I wrote it is too broad.

abhishekbose87:
I guess you are saying that `Person` is commentable. Let me explain what I thought.
I was thinking that items, tasks will be commentable but those comments will be made by Person object. Does that make sense or did I totally go on the wrong track.

> What you can do is write a migration to move the existing comments to something like old _ comments. Ideally the migration would move the old into the new format

Could I remove the field altogether unless there is live data which we don't want to lose.

> Perhaps you should get this working as polymorphic, then as a concern? I'll happily accept two PRs. The scope of the issue as I wrote it is too broad.

One PR should be good. As you said, the scope was a little broad so I needed to get some feedback. But yeah we should be fine.

kgf:
Sadly there is some live data, especially on the Person model that I would like to retain.
Person, Item et al are alike, they are `Commentable`

Does that help ?
Thanks
Kevin

AliaTabassumAS:
.

abhishekbose87:
@kgf Thanks for the info 👍 

I will update the PR by tomorrow. Let you know if I have more questions.

abhishekbose87:
Hey @kgf, Can you have a look now and provide feedback.

Btw I really made heavy work at the end, by rebasing my branch with `master` (and it took so much time). After which I realised that `development` was the correct branch :facepalm:

app/views/people/_form.html.erb
```diff
@@ -58,6 +58,6 @@
     <h4> Personnel Image </h4>
     <%= @person.portrait %>
     <%= f.file_field :portrait %>
-    <%= f.input :comments, :input_html => { :size => 40} %>
+    <%= f.input :old_comments, :input_html => { :size => 40} %>
```
kgf, line 61: Seems like you'll also need to add a table to display the new comments ?

spec/factories/comments.rb
```diff
@@ -0,0 +1,5 @@
+FactoryGirl.define do
+  factory :comment do
+
```
kgf, line 3: It would be good to have a valid factory

kgf:
Thanks for all your help. Sorry to hear about the difficulties with merging. 

I made a couple of comments. I don't know if you have time to address them, or should I create new issues for them ?

PR #281, opened 2016-11-01
Add Notifications tab to Event page

### LINKED ISSUE #275: Add data table to Events for Notifications
On the event show page, there are currently two tabs. This would be to add an additional tab. The tab should have a data table showing notifications for this event. You should use the same JS code used for the Availability listing on one of the other tabs.

jeffhos:
Moved Notifications table into a partial and use it in the Event display.  Resolves #275.

### DIFF
--- a/app/views/notifications/_notifications_table.html.erb
+++ b/app/views/notifications/_notifications_table.html.erb
@@ -11,7 +11,8 @@
       <th colspan="1"></th>
     </tr>
   </thead>
-I need to allow info only Notifications
+
+  <%# TODO: I need to allow info only Notifications %>
   <tbody>
     <% notifications.each do |notification| %>
       <tr>
--- a/app/views/notifications/_notifications_table.html.erb
+++ b/app/views/notifications/_notifications_table.html.erb
@@ -1,0 +1,26 @@
+<table id="notifications" class="generic_datatable table table-striped table-bordered">
+  <% if defined? title %>
+    <caption><h3><%= title %></h3></caption>
+  <% end %>
+  <thead>
+    <tr>
+      <th>Status</th>
+      <th>Event</th>
+      <th>Start at</th>
+      <th>Channels</th>
+      <th colspan="1"></th>
+    </tr>
+  </thead>
+I need to allow info only Notifications
+  <tbody>
+    <% notifications.each do |notification| %>
+      <tr>
+        <td><%= notification.status %></td>
+        <td><%= notification.event ? (link_to notification.event.title, notification.event) : 'None' %></td>
+        <td><%= notification.scheduled_start_time %></td>
+        <td><%= notification.channels %></td>
+        <td><%= link_to 'Edit', edit_notification_path(notification) %></td>
+      </tr>
+    <% end %>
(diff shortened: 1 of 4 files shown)


kgf:
Thanks for your help. I'll review this later today.

app/views/notifications/_notifications_table.html.erb
```diff
@@ -0,0 +1,27 @@
...
+
+  <%# TODO: I need to allow info only Notifications %>
+  <tbody>
+    <% notifications.each do |notification| %>
+      <tr>
+        <td><%= notification.status %></td>
+        <td><%= notification.event ? (link_to notification.event.title, notification.event) : 'None' %></td>
+        <td><%= notification.scheduled_start_time %></td>
+        <td><%= notification.channels %></td>
+        <td><%= link_to 'Edit', edit_notification_path(notification) %></td>
```
kgf: Could you please add the CanCan permission here ? `if can edit @notification`

app/views/notifications/_notifications_table.html.erb
```diff
@@ -0,0 +1,27 @@
...
+      <th>Channels</th>
+      <th colspan="1"></th>
+    </tr>
+  </thead>
+
+  <%# TODO: I need to allow info only Notifications %>
+  <tbody>
+    <% notifications.each do |notification| %>
+      <tr>
+        <td><%= notification.status %></td>
```
kgf, line 19: We should add a column that is a truncation of the subject. Also, that column should be a link to the notification itself.

kgf:
Looks pretty good, few small changes. 
If you had time to add some tests, that would also be great.

Thanks for your help,
Kevin

app/views/notifications/_notifications_table.html.erb
kgf: Here is an example
 https://github.com/ReadyResponder/ReadyResponder/blob/development/app/views/people/index.html.erb#L39

app/views/notifications/_notifications_table.html.erb
jeffhos: I don't see `subject` or anything that looks like that on `notification`.  What attribute did you want to display here?

kgf: Hey, aesome, it's not there.
It is here however
https://github.com/ReadyResponder/ReadyResponder/wiki/Notifications
Skip that then, I'll make a new issue for it directly, beyond the scope of this one.
Thanks !
Kevin

app/views/notifications/_notifications_table.html.erb
kgf: If you could still get this one, I think we'd be ready to merge.

kgf:
I don't quickly see why the tests have failed, but the failure is related to this pr

jeffhos:
Yup, it has to do with adding the permission test to the notifications page... I'm looking at it right now.

kgf:
Thanks for seeing this through to completion.

jeffhos:
No problem... sorry it took a bit to get it sorted out.

PR #284, opened 2016-11-02
262 notification statuses

amandamunoz:


app/models/notification.rb
```diff
@@ -2,8 +2,22 @@ class Notification < ActiveRecord::Base
   belongs_to :event
   # has_many :recipients
 
-  VALID_STATUSES = %w(pending active canceled complete expired)
-  validates :status, inclusion: { in: VALID_STATUSES }
```
kgf, line 6: Should we keep the validation ?

PR #162, opened 2016-10-04
Grant Model

pcofilada:
Added Grant Model and associate Item model on it.

PR #286, opened 2016-11-04
Pcofilada feature/grant model

kgf:
These are the changes that @pcofilada did not have time to complete.

See #162


kgf:
I will need to deal with these test failures later.

kgf:
This has far too much excess baggage. Will start a new PR from dev.

PR #288, opened 2016-11-08
Add body and subject to notifications

romangurovich:
Let me know if we want any validations. Right now, anything goes.


spec/factories/notifications.rb
```diff
@@ -1,5 +1,7 @@
 FactoryGirl.define do
   factory :notification do
+    subject "Urgent: Read Me"
+    body "We live in a vacuous world, yet we do so with a feeling of urgency."
```
kgf, line 4: Nice, I like good factories !

kgf:
Thanks for your help !

PR #291, opened 2016-11-20
Upgrade rails from 4.2.7 to 4.2.7.1.

houhoulis:
As pointed out by the Snyk service.

`bundle update rails` made the tiny version update to rails, but also brought along seemingly unrelated updates:

* rack - supports a minor new feature
* nokogiri - It's a tiny version change, but it removes the `pkg-config` dependency as optional, which sounds like a *big* change but what do I know.
* minitest
* rake

I don't think any of these additional upgrades are required, but I'm not aware of a reason to exclude them.  (Note that rake, nokogiri, and minitest aren't explicitly required in our Gemfile -- they're dependencies.)


kgf:
Thanks for your help, Chris

PR #292, opened 2016-11-26
200 send messages via text

kgf:
Removed some of the service objects, and moved the methods into the model classes. Also got back to green.


Gemfile
```diff
@@ -1,4 +1,4 @@
-source 'https://rubygems.org'
+source 'http://rubygems.org'
```
houhoulis, line 1: I'm not sure whether this change is inadvertent, or a workaround for something (perhaps the Rubygems SSL change?), or ???

But, I don't think it's appropriate to remove SSL protection, is it?

kgf: I believe this was a Bundler issue. I will happily revisit re-adding it back in the future.

houhoulis: 😕

houhoulis:
Some inconsistencies were introduced in this PR, as well as other PRs recently, that will take significant time to repair.  I'm writing here to warn about merge commits.

Whenever a PR includes a merge commit, there is a chance that changes will be included that are:

1. unintended (because of the size of many merge commits)
1. hard to resolve manually, because changes are hard to trace past the merge commit back to their origins

In my experience, when the merge commit is a merge from development into the topic branch, and then that branch is merged back into development, any inadvertent changes are trickier than most to pull back out, even after 1. and 2. above.  (Although 1., the size, is often the biggest obstacle.)  This is because git seems to figure it should resolve any merge conflicts silently to follow what was done during the round-trip merges.

Different projects handle merge conflicts and shared history in git differently.  I prefer (and recommend) to never merge a merge commit into the common codeline, instead relying on rebasing, cherry-picking, or other techniques.  The inconsistencies that were introduced in this PR is why.

This PR included inadvertent or inadvisable changes to db/schema.rb.  Changes to db/schema.rb that are inconsistent with a project's existing migrations have a tendency to spread.  They cause an accelerating number of merge conflicts among developers, and it becomes hard to figure out the true current state of the database.  This PR includes:

* Many, many lines that define new constraints on database fields, without a migration to accomplish such constraints.
* A "grants" table and "grant_id" field(s) with no migration to accomplish the change.
  * There's also no definition of a Grant model.
  * `Item` now has both a `grant` and a `grant_id` field.  This means that if we add a Grant model,  `item.grant` could (maybe?) be a different class of object for different developers.
* It's possible that there are other changes - which are hard to detect due to the many, many lines that newly constrain database fields.

This PR also included many other changes that are unrelated to "send messages via text".  Service objects (great!), specs to enforce changes in functionality across the app, enhancements to spec factories, a cosmetic change in the Item index view, removing some values from a class constant in Channel, ... The large number of changes in the PR decreased the chance that someone could promptly afford the time necessary to properly review them.

Hopefully someone can follow up soon to fix the inconsistencies -- perhaps myself.  But just looking into all these changes has already taken a bunch of time.

PR #295, opened 2016-11-30
Broadcast notifications

kgf:
Ongoing work, but smaller PRs get merged.


app/views/recipients/edit.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1>Editing Recipient</h1>
+
+<%= render 'form' %>
+
+<%= link_to 'Show', @recipient %> |
+<%= link_to 'Back', recipients_path %>
```
david135, line 6: Should we use the `content_for :sidebar` approach here?

app/views/recipients/new.html.erb
```diff
@@ -0,0 +1,5 @@
+<h1>New Recipient</h1>
```
david135, line 1: Should we use the `content_for :sidebar` approach here?

app/views/recipients/show.html.erb
```diff
@@ -0,0 +1,4 @@
+<p id="notice"><%= notice %></p>
```
david135, line 1: Should we use the `content_for :sidebar` approach here?

kgf: I agree, and I have created an issue #296

PR #297, opened 2016-12-07
Hotfixes that were needed to run in production

kgf:
There was also a change to config/devise that
will not be commited into the repo.


kgf:
Opening this from my fork

Issue #231, opened 2016-10-23
Make top menu dynamic for people


kgf:
Right now, the People menu drops down to reveal sub menus. The first two choices are currently static. They should be changed to reflect the Active departments that have the `We manage People` setting set to true. You should use `shortname` attribute for display in the menu. This will also require changes in the people controller, since the values are also hard-coded there.

sergeimellow:
taking a look at this now

kgf:
Awesome, let me know if you have questions !

sergeimellow:
@kgf Is there a reason that the relationships aren't established between departments and people? Person.first.department just returns a string instead of the department the user is related to.

kgf:
Almost certainly an oversite from a time when Departments weren't a model. At this point, that would be a bug !

sergeimellow:
@kgf This will help the next person who tries to tackle the initial problem whether its me or somebody else. I am done for now.

https://github.com/ReadyResponder/ReadyResponder/pull/242/files

kgf:
Thanks for your help ! I appreciate it.

kgf:
Resolved in #298

PR #298, opened 2016-12-07
Make top menu dynamic for people

pcai:
proof of concept for #231 - dynamic "People" submenu that is populated with departments that manage people. Replaces the hardcoded CERT and Police department routes.

There are still many places where department is treated as a string, but I thought I'd open this PR early to check if I understand the feature request.

UPDATE:
As requested, this PR also transitions to required department shortnames. The included migration should handle most cases without breaking anyone, and the seed.rb file has been updated accordingly.

capture

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -27,16 +27,10 @@ class PeopleController < ApplicationController
     render :template => "people/index"
   end
 
-  def police
-    redirect_to department_people_path("Police")
-  end
   def everybody
     @people = Person
     @page_title = "Everybody"
     render :template => "people/index"
-  end
-  def cert
-    redirect_to department_people_path("CERT")
   end
 
   def other
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -16,11 +16,19 @@ class PeopleController < ApplicationController
     @page_title = "Org Chart"
     render :layout => "orgchart"
   end
+  
+  def department
+    dept = Department.find(params[:dept_id])
+    # TODO: more gracefully handle department not found
+    head 404 and return if dept.nil?
+    
+    @people = Person.active.where(department_id: dept.id)
+    @page_title = dept.name
+    render :template => "people/index"
+  end
 
   def police
-    @people = Person.active.select{|person| person.department&.name == "Police"}
-    @page_title = "Police"
-    render :template => "people/index"
+    redirect_to department_people_path("Police")
   end
   def everybody
     @people = Person
@@ -28,9 +36,7 @@ class PeopleController < ApplicationController
     render :template => "people/index"
   end
   def cert
-    @people = Person.active.select{|person| person.department&.name == "CERT"} + Person.cert.all
-    @page_title = "CERT"
-    render :template => "people/index"
+    redirect_to department_people_path("CERT")
   end
 
   def other
--- a/app/models/department.rb
+++ b/app/models/department.rb
@@ -8,6 +8,10 @@ class Department < ActiveRecord::Base
   has_many :items
   has_many :locations
   has_and_belongs_to_many :notifications
+  
+  validates_presence_of :shortname
+  
+  scope :active, -> { where( status: "Active" ) }
 
 
   def self.managing_people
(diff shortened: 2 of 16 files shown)


app/views/layouts/_top_menu.html.erb
```diff
@@ -15,8 +15,9 @@
           <li class="dropdown">
             <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button">People<span class="caret"></span></a>
             <ul class="dropdown-menu">
-              <%= nav_link "Police", police_people_path %>
-              <%= nav_link "CERT", cert_people_path %>
+              <% Department.managing_people.order("shortname ASC").each do |dept| %>
+                <%= nav_link dept.shortname, department_people_path(dept) %>
```
pcai, line 19: unfortunately, this has the effect of omitting otherwise valid departments, if the department shortname is nil/empty.

app/controllers/people_controller.rb
```diff
@@ -16,21 +16,27 @@ def orgchart
...
+    @people = Person.active.where(department_id: dept.id)
+    @page_title = dept.name
+    render :template => "people/index"
+  end
 
   def police
-    @people = Person.active.select{|person| person.department&.name == "Police"}
-    @page_title = "Police"
-    render :template => "people/index"
+    redirect_to department_people_path("Police")
```
pcai: this doesnt work, as the parameter to this route is an id, not a shortname as I originally intended. I thought the shortnames would work fine, but since shortnames are mutable it felt a bit weird that changing a shortname would break links to a specific department's people page.

app/views/layouts/_top_menu.html.erb
kgf: I am perfectly happy with it being a required field.

pcai: Any advice on how to do this? I can add a migration to enforce it at the db level, but it would rollback if any existing rows would fail the new constraint. Should I copy a default value over from the name as part of the migration, if necessary? Or would the policy be to operate as if all the data was already valid? I think even the seeds.rb file for example neglects adding shortname.

app/controllers/people_controller.rb
kgf: I think we'll need to break more here. There can't be a `police` method, it won't scale. If this is just for backwards compatibility, go ahead and take it out. Production users have not created enough bookmarks that I'm concerned about it,
Thanks for your help !

app/views/layouts/_top_menu.html.erb
kgf: We'll require a short name.

config/routes.rb
```diff
@@ -75,6 +75,7 @@
       get 'signin'
       get 'orgchart'
       get 'roster'
+      get 'department/:dept_id', action: "department", as: :department
```
kgf, line 76: We can drop the police and CERT routes above, too.

app/views/layouts/_top_menu.html.erb
kgf: Not sure how my comment duplicated. 
I am not aware of any production deployments that don't have short name in the departments.
Seeds should be updated.
I really like the idea of the default shortname in the migration !

pcai:
This should be ready for another look, assuming travis passes

app/models/department.rb
```diff
@@ -8,6 +8,10 @@ class Department < ActiveRecord::Base
   has_many :items
   has_many :locations
   has_and_belongs_to_many :notifications
+  
+  validates_presence_of :shortname
+  
+  scope :active, -> { where( status: "Active" ) }
 
 
   def self.managing_people
```
kgf, line 17: Not in the scope of this PR, but should this be a scope?

pcai: Probably? Right now it wouldn't be possible to chain `managing_items` with `managing_departments` for example. And it's more idiomatic. But otherwise not much benefit I suppose.

kgf:
This is great, thanks. I'm ready to merge after #299

pcai:
Glad I could help. I'll fix the conflicts after that merges.

app/models/department.rb
kgf: Sometimes idiomatic is good. Since I like to encourage junior developers to join us, don't want to teach bad habits !

kgf:
Would you have time to look at that PR ? Also, if you're on Slack, we have a team there, I can invite you.

pcai:
@kgf thanks - I'd be happy to join the slack. My email is my first name at privy

kgf (approved):
This looks good, removes some of the hard coded references.
Awaiting #299 to be merged

pcai:
Conflicts fixed. Not sure why codacy is now complaining about a line that didn't change  ¯\\\_(ツ)_/¯

PR #299, opened 2016-12-08
Hotfixes that were needed to run in production

kgf:
There was also a change to config/devise that
will not be commited into the repo.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -71,6 +71,7 @@ class PeopleController < ApplicationController
   end
 
   def index
+    #Person.active.joins(:department).where("departments.manage_people": true)
     @people = Person.active.select{|person| person.department&.shortname == 'CERT' || person.department&.shortname == 'BAUX'}
     @page_title = "Active Police and CERT"
     respond_to do |format|
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -5,15 +5,15 @@ class PeopleController < ApplicationController
 
   def signin
     #This is the sign-in sheet, not anything about authentication
-    @department = Department.where(shortname: "BAUX")&.first
-    @people = @department.people.active
+    @department = Department.where(shortname: "BAUX").first
+    @people = @department.people.active if @department
     @page_title = "Sign-in"
     render :layout => "print_signin"
   end
 
   def orgchart
-    @department = Department.where(shortname: "BAUX")&.first
-    @people = @department.people.active
+    @department = Department.where(shortname: "BAUX").first
+    @people = @department.people.active if @department
     @page_title = "Org Chart"
     render :layout => "orgchart"
   end
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -18,7 +18,8 @@ class PeopleController < ApplicationController
   end
 
   def police
-    @people = Person.active.select{|person| person.department&.name == "Police"}
+    @department = Department.where(shortname: "BAUX").first
+    @people = @department.people.active
     @page_title = "Police"
     render :template => "people/index"
   end
@@ -28,7 +29,8 @@ class PeopleController < ApplicationController
     render :template => "people/index"
   end
   def cert
-    @people = Person.active.select{|person| person.department&.name == "CERT"} + Person.cert.all
+    @department = Department.where(shortname: "CERT").first
+    @people = @department.people.active
     @page_title = "CERT"
     render :template => "people/index"
   end
@@ -68,7 +70,7 @@ class PeopleController < ApplicationController
   end
 
   def index
-    @people = Person.active.select{|person| person.department&.name == 'CERT' || person.department&.name == 'Police'}
+    @people = Person.active.select{|person| person.department&.shortname == 'CERT' || person.department&.shortname == 'BAUX'}
     @page_title = "Active Police and CERT"
     respond_to do |format|
       format.html # index.html.erb
@@ -91,7 +93,6 @@ class PeopleController < ApplicationController
     @person = Person.new(status: cookies[:status], state: 'MA')
     @person.emails.build(category: 'E-Mail', status: 'OK', usage: '1-All')
     @person.phones.build(category: "Mobile Phone", status: "OK", usage: "1-All")
-
     respond_to do |format|
       format.html # new.html.erb
       format.json { render json: @person }
@@ -101,6 +102,8 @@ class PeopleController < ApplicationController
   def edit
     @person = Person.includes(:phones, :emails).find(params[:id])
     @page_title = @person.fullname
+    @emails = @person.emails
+    @phone = @person.channels.select {|p| p.phone?}.first
   end
 
   def create
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -5,14 +5,15 @@ class PeopleController < ApplicationController
 
   def signin
     #This is the sign-in sheet, not anything about authentication
-    @people = Person.active
-    @department = Department.where(name: "Police")&.first
+    @department = Department.where(shortname: "BAUX")&.first
+    @people = @department.people.active
     @page_title = "Sign-in"
     render :layout => "print_signin"
   end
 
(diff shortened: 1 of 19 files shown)


app/controllers/people_controller.rb
```diff
@@ -5,20 +5,22 @@ class PeopleController < ApplicationController
 
   def signin
     #This is the sign-in sheet, not anything about authentication
-    @people = Person.active
-    @department = Department.where(name: "Police")&.first
+    @department = Department.where(shortname: "BAUX")&.first
```
david135: Here and on line 15, the safe navigation operator may not help much, since the next line will fail if it isn't set properly.  Better, I think, to fail here than on the next line (as in lines 22 and 33).

app/controllers/people_controller.rb
```diff
@@ -68,7 +71,7 @@ def inactive
   end
 
   def index
-    @people = Person.active.select{|person| person.department&.name == 'CERT' || person.department&.name == 'Police'}
+    @people = Person.active.select{|person| person.department&.shortname == 'CERT' || person.department&.shortname == 'BAUX'}
```
pcai, line 75: I believe currently this loads all active people, doing the filtering in ruby code instead of at the database level. There may also be an n+1 query problem. See the queries produced for 3 matching people:

```
Person.active.select{|person| person.department&.shortname == 'CERT' || person.department&.shortname == 'BAUX'}
  Person Load (0.6ms)  SELECT "people".* FROM "people" WHERE "people"."status" = $1  ORDER BY division1, division2, title_order, start_date ASC  [["status", "Active"]]
  Department Load (0.3ms)  SELECT  "departments".* FROM "departments" WHERE "departments"."id" = $1 LIMIT 1  [["id", 4]]
  Department Load (0.3ms)  SELECT  "departments".* FROM "departments" WHERE "departments"."id" = $1 LIMIT 1  [["id", 4]]
  Department Load (0.3ms)  SELECT  "departments".* FROM "departments" WHERE "departments"."id" = $1 LIMIT 1  [["id", 4]]
  Department Load (0.3ms)  SELECT  "departments".* FROM "departments" WHERE "departments"."id" = $1 LIMIT 1  [["id", 5]]
```

would this be better ?:

`Person.active.joins(:department).where("departments.shortname": ['CERT', 'BAUX'])`

or to avoid the join entirely by using two queries:

`Person.active.where(department_id: Department.where(shortname: ['CERT', 'BAUX']).pluck(:id))`

pcai: the two query solution or one involving a subquery would also be nicer as it can use the department_id index.

app/controllers/people_controller.rb
kgf: Actually, this will never fail, it will just return nil. So, yes, better to remove the `&`

app/controllers/people_controller.rb
kgf: Those are better solutions, but I must confess I did not give this my full attention. I believe this all goes away with your next PR ?

houhoulis: What PR is the "next PR"?

houhoulis: I agree with @pcai that the query in this PR

1. is an "n + 1" query, which is non-ideal
1. filters by instantiating all People in Ruby, which is non-ideal

I like Peter's first version of the query best.  Whether the second version makes use of that index is beyond my intuition (thanks @pcai , I will look into it further!)

Another possibility is: since the view instantiates each user's department to get its name, maybe Peter's second query should be altered by the addition of `includes(:department)`, and would then perform best.

I think the real-world difference in expected performance among these suggestions is pretty small.  However, in the real world, they are *all* much more performant (and better practice) than the query in this PR.

app/controllers/people_controller.rb
```diff
@@ -5,20 +5,22 @@ class PeopleController < ApplicationController
 
   def signin
     #This is the sign-in sheet, not anything about authentication
-    @people = Person.active
-    @department = Department.where(name: "Police")&.first
+    @department = Department.where(shortname: "BAUX").first
```
houhoulis, line 8: This PR continues to hard-code strings into ReadyResponder ... including into the `signin`, `orgchart`, and `index` methods.  Would now be a good time to, instead, get away from such a practice?

I would have suggested considering whether this query should instead be a scope in the Person model, except that I'm unclear why we are hard-coding this query at all.

app/views/layouts/_footer.html.erb
```diff
@@ -1,5 +1,5 @@
-Version 0.81
+Version 1.9.1
```
houhoulis: Hard-coded?

spec/features/person_displays_spec.rb
```diff
@@ -126,13 +126,18 @@
...
       find('#navbar').click_link('Other')
       #visit other_people_path
       expect(page).to have_content('Others')
       expect(page).to have_content('Home') # This is in the nav bar
       # Factory Gem Creates a Second CJ that isn't a police officer or CERT so this part fails.
       # The original CJ passes with flying colors and is not displayed
+
+      # KGF 12/7/16 I wonder if this is related to my issue at line ~33
+      # I think this entire block needs to go. There probably won't be an 'other'
+      # TODO reenable this assertion
```
houhoulis, line 140: This shouldn't be in a PR into upstream.

app/services/msg/upcoming.rb
```diff
@@ -3,14 +3,14 @@ def respond
     events = Event.upcoming.limit(5)
     response = ""
     events.each do |event|
-      response += "Start #{event.start_time.to_s} " if event.start_time.present?
-      response += "End #{event.end_time.to_s} " if event.end_time.present?
+      response += "#{event.start_time.strftime('%a %b %d %k:%M')} - " if event.start_time.present?
+      response += "#{event.end_time.strftime('%b %d %k:%M')} " if event.end_time.present?
       response += " [#{event.id_code.to_s}] " if event.id_code.present?
-      response += event.title.truncate(15) if event.title.present?
+      response += event.title.truncate(23) if event.title.present?
```
houhoulis, line 9: A better design might be to define this `response` string in a presenter method on an `Event`.  (whether it's an actual Decorator/Presenter class, an Event service object, or an Event instance method, all are reasonable.)  Then this entire service object could be replaced by, at most,

```ruby
   Event.upcoming.limit(5).map(&:response).join("\n")
```

And since this entire service object is about Events, the above line of code could be in an Event Presenter, an Event service object, or an Event class method.

app/views/layouts/_footer.html.erb
kgf: Currently, yes. Don't have all the niceties yet.

spec/features/person_displays_spec.rb
kgf: What would you prefer? I don't want to overlook that this assertion has been commented out.

app/controllers/people_controller.rb
kgf: It would actually not be a good time, because this PR should not have any refactoring in it. This PR does not represent what I want development to be, but instead represents what Production *is*. I needed to make these changes, or the app would not function.
However, I am *very* anxious to remove these hard-coded references, and support any effort towards that. Just not in _this_ PR. I want to have a working point I can tag.

app/controllers/people_controller.rb
kgf: How about
```
    Person.active.joins(:department).where("departments.manage_people": true)
```
If we're going to refactor it, let's get the hard coded strings out.

pcai:
Since this branch is required to run in production, I am happy to have this merge as-is, and I can take a look at fixing the hardcoding issues that I've started working on in https://github.com/ReadyResponder/ReadyResponder/pull/298

kgf:
Thanks @pcai , I have a version in which the tests should pass. It includes a comment on an approach to removing the hard coded references.

kgf:
This passes the test suite.

kgf:
@pcai, when you have a chance, if you could pull the trigger on this. Passing the test suite.
Thanks !
Kevin

pcai (approved):

Issue #290, opened 2016-11-14
Move action buttons to sidebar


kgf:
All actions should be in the sidebar, rather than at the bottom.

In this file 
app/views/settings/show.html.erb
and this
app/views/settings/new.html.erb

And anywhere else for settings

anncolvin:
I'd love to take this one!

david135:
Great!  Thanks for taking this on, @anncolvin.

kgf:
Yes, thanks @anncolvin . Let us know if you have any questions.

kgf:
@anncolvin Are you having any success ? Do you have questions ?

anncolvin:
I just submitted a pull request. Please let me know if there are any issues with it.

kgf:
Closed by #302 
Thanks for your help @anncolvin

PR #302, opened 2016-12-12
moved actions in settings to sidebar, issue #290

anncolvin:


kgf (approved):
This is a good start. We'll need a second issue to have the buttons match others, such as having `Cancel` be a button during editing.
This can be merged after #299

PR #303, opened 2016-12-20
31 link departments and events

kgf:
This allows an event to have only the correct people, not everyone.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -4,7 +4,8 @@ class Event < ActiveRecord::Base
 
   attr_accessible :title, :description, :category, :course_id, :is_template,
                   :duration, :start_time, :end_time, :instructor, :location,
-                  :id_code, :status, :timecard_ids, :person_ids, :comments
+                  :id_code, :status, :timecard_ids, :person_ids, :comments,
+                  :department_ids
 
   validates_presence_of :category, :title, :status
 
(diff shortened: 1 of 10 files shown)


app/models/event.rb
```diff
@@ -64,12 +66,11 @@ def respondents
   end
 
   def eligible_people
-    Person.active.all # In the future, this will need to honor department
-    # def eligible_people
-    #   self.departments.each do |department|
-    #
-    #   end
-    # end
+    eligible_people = []
```
pcai: may i suggest this instead?

`return Person.active.where(department_id: self.department_ids)`

If `department_ids` does what I think it does, and all the data we want is already in the database, then this should query the precise set of people we care about, and as a bonus it will be a lazy loaded `ActiveRecord::Relation`, instead of being instantiated as an array of people immediately.

This would not work though if there were any changes to the current Event's department associations that have not yet been saved to postgres.

kgf: I'm ok with that edge case, which I think would also be present in my implementation. 
I played around a whlie trying to get this into a one liner, but kept coming at it from the event, not Person. Thanks ! I'm going to go try it.

kgf: Even slightly better
`Person.active.where(department: self.departments)`

pcai: It would be very hard from the perspective of the event, as you have to "hop" across two tables - the new join table and then from department to person. And I think rails doesn't give you the handy `has_many :through` construct when that's the case.

kgf: The `has many through` is the other variant of `has and belongs to many`, which I used here. I did not need any additional info on the join table.

Issue #307, opened 2017-01-04
db/schema.rb is out of sync with migrations again


houhoulis:
We haven't yet located and committed all the migrations necessary to produce the currently-checked-in `db/schema.rb`.

This means that people with existing schemas will find local changes once they run migrations to catch up to recent changes on development.

Here are all the things that are missing between the current migrations and the current `db/schema.rb` :

```diff
diff --git a/db/schema.rb b/db/schema.rb
index 3dc7b14..b2fc3e3 100644
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -17,12 +17,12 @@
   create_table "activities", force: :cascade do |t|
-    t.string   "content"
-    t.string   "author"
+    t.string   "content",       limit: 255
+    t.string   "author",        limit: 255
     t.integer  "loggable_id"
-    t.string   "loggable_type"
-    t.datetime "created_at"
-    t.datetime "updated_at"
+    t.string   "loggable_type", limit: 255
+    t.datetime "created_at",                null: false
+    t.datetime "updated_at",                null: false
   end
 
   create_table "availabilities", force: :cascade do |t|
@@ -40,39 +40,39 @@
   create_table "certs", force: :cascade do |t|
     t.integer  "person_id"
     t.integer  "course_id"
-    t.string   "status"
-    t.string   "category"
-    t.string   "level"
-    t.string   "cert_number"
+    t.string   "status",          limit: 255
+    t.string   "category",        limit: 255
+    t.string   "level",           limit: 255
+    t.string   "cert_number",     limit: 255
     t.date     "issued_date"
     t.date     "expiration_date"
     t.text     "comments"
     t.integer  "updated_by"
     t.integer  "created_by"
-    t.datetime "created_at"
-    t.datetime "updated_at"
-    t.string   "grade"
+    t.datetime "created_at",                  null: false
+    t.datetime "updated_at",                  null: false
+    t.string   "grade",           limit: 255
     t.integer  "event_id"
-    t.string   "certification"
+    t.string   "certification",   limit: 255
   end
 
   add_index "certs", ["person_id"], name: "index_certs_on_person_id", using: :btree
 
   create_table "channels", force: :cascade do |t|
     t.integer  "person_id"
-    t.string   "name"
-    t.string   "status"
-    t.string   "content"
+    t.string   "name",          limit: 255
+    t.string   "status",        limit: 255
+    t.string   "content",       limit: 255
     t.integer  "priority"
-    t.string   "category"
[...]

PR #308, opened 2017-01-04
Sending notifications.

kgf:
Still need to save the message as a record of what was sent.
This is so others can comment.


kgf:
This may also addres #307 as I rebuilt my db from scratch.

PR #309, opened 2017-01-05
added Msg::Base#get_event_codename

masukomi:
* adds a helper method to make it easy and obvious to extract event codenames
from params.
* refactored available and unavailable to use it
* added spec for new method


kgf:
Thanks @masukomi

PR #315, opened 2017-01-14
Adds support for Markdown based helpdocs.

masukomi:
As discussed in the Slack channel. this uses the `markdown_rails` gem and `RedCarpet` for rendering. I've enabled support for fenced codeblocks and tables in the initializer if we want them.

I've removed support for rendering them via json because i don't actually know what that would realistically be. 

usage is pretty simple: just add `foo.html.md` to `app/views/helpdocs/` then request `/helpdocs/foo` and it'll be rendered.

We can make these .html.md.erb files, but right now we don't know what variables we need to be loading.


config/routes.rb
```diff
@@ -38,6 +38,7 @@
 
   get 'landing/help', as: 'help', path: '/help'
 
+
```
kgf: Extra ?

app/controllers/helpdocs_controller.rb
```diff
@@ -1,83 +1,12 @@
...
 
   # GET /helpdocs/1
   # GET /helpdocs/1.json
   def show
-    @helpdoc = Helpdoc.find(params[:id])
 
     respond_to do |format|
-      format.html # show.html.erb
-      format.json { render json: @helpdoc }
+      format.html { render params[:id] }
```
kgf: Does this Codacy issue concern you ?

Gemfile
```diff
@@ -8,11 +8,14 @@ gem 'carrierwave'
...
 gem 'rmagick', '~> 2.15.4'
 gem 'simple_form'
+gem 'markdown-rails'
+gem 'redcarpet'
 gem 'geocoder'
 gem 'cancancan'
 #gem "taps"
 #gem 'validates_timeliness', '~> 3.0'
 
+
```
kgf, line 17: Extra Newline ?

masukomi:
Ok. I'm done poking this. It should be good to go. 

## Notes: 
The 'helpdocs/index' is mostly intended for internal use at the moment. I think it should have some beautification before we link to it from anywhere. I'm thinking extract the first title from the md file and use that as the name that gets used in the link? 

I've changed the id of the tasks datatable because now the markdown is generating ids for every header and the jQuery matcher was looking for anything, on any page, with the id of "tasks" which is just WAY too generic. I made it "tasks_table" but really the datatable matchers really need to be looking for much more specific things, not just any element on any part of any page with a matching name.

I've punted on the controller tests until we have helpdocs to test. As soon as there's something in the system we can write a test to confirm that some generic aspect of it is being rendered correctly. Right now even the index page is just an empty list (because there are no help files other than itself) and I'm not sure if the title at the top is what we want.

masukomi:
fixed the helpdocs related failures in travis. I have no clue why notifications was failing but I don't see how it has anything to do with this pr. 

addressed the codacity formatting and security concerns. I added in a path check to see if the thing actually existed. technically a well crafter url COULD result in rendering some other `<something>.html.md` file but i don't _think_ that's  a huge concern because the only .md files are (currently) helpdocs and anything else would be almost guaranteed to have same erb params in it that would be missing and cause it to blow up. if it didn't that would mean it was some random unparameterized md template and ..... unless you're storing secrets in the deployed files in .html.md templates.... that should be just fine.

masukomi:
i'm failing the following test. I don't see how this has anything to do with my PR. I suspect it's a bug in the main repo. 

```
  1) Notifications  visit notifications from an event should create a notification for that event
     Failure/Error:
       within("#flash_notice") do
         expect(page).to have_content "Notification was successfully created."
       end
     
     Capybara::ElementNotFound:
       Unable to find css "#flash_notice"
     # ./spec/features/notifications_spec.rb:20:in `block (4 levels) in <top (required)>'
     # ./spec/support/database_cleaner.rb:10:in `block (2 levels) in <top (required)>'
```

masukomi:
also, i'm not sure what to do about the codacity  issue because google isn't showing me how to tell it to ignore a fingerprint like code climate

pcai:
I believe the codacy issue is legitimate, because this method may introduce a potential directory traversal exploit.

For example, I was able to navigate to `/helpdocs/%2E%2E%2Fevents%2Fshow` which urldecodes to `/helpdocs/../events/show`. 

This causes `path` to be set to:
`#<Pathname:/home/pcai/lims/app/views/events/show.html.md>` - which luckily, is not a valid file, because the extension is wrong (its .erb instead of .md). But I think it has the potential to be combined with another flaw in the future.

May I suggest a change like this perhaps? 
```ruby
file_name = File.basename(params[:id])
render file_name
```
http://ruby-doc.org/core-1.9.3/File.html#method-c-basename

Update: I also opened a PR for the notification spec failure.

kgf:
@masukomi I had the same spec issue regarding notifications. @pcai reports that may be fixed in #325

masukomi:
I ended up with a solution that was a little of mine, a little of @pcai 's Hopefully it'll make Codacity happy.

config/routes.rb
```diff
@@ -10,7 +10,7 @@
   resources :resource_types
   resources :messages
   resources :activities
-  resources :helpdocs
+  resources :helpdocs, only: [:show, :index]
```
kgf, line 13: I like this

app/controllers/helpdocs_controller.rb
```diff
@@ -1,83 +1,31 @@
...
-        format.json { head :no_content }
-      else
-        format.html { render action: "edit" }
-        format.json { render json: @helpdoc.errors, status: :unprocessable_entity }
+    file_name = "#{File.basename(params[:id])}.md"
+    path = Rails.root.join('app', 'views', 'helpdocs', file_name)
+    if File.exist? path
+      respond_to do |format|
+        format.html {
+          render file_name
```
kgf, line 25: How does this activate red carpet to parse the mark down?

masukomi: it works the same way as if it were a haml file. You create a template handler and register it, tell it what extensions it supports ( see `config/initializers/markdown_handler.rb`)  and the render method figures it out from there. What I can't figure out why (in this code) `.md` works but `.html.md.erb` does not. I thought I'd be able to just register a handler for the pipeline and it would handle things one extension at a time.... but no. If i want it to handle the ERB code in it at the same time it has to be `.md` not `.md.erb` and i just have to override the `erb` method in my handler. Doesn't seem right but...

masukomi:
codacity didn't like that either. I think with @pcai 's suggestion there's no way this is a real security risk.

pcai:
I agree there is no security risk now. Upon further reading of the documentation in codacity, this is probably an unavoidable false positive:

> This warning is shown whenever the path to be rendered is not a static string or symbol. 

So oh well. Looks 👍 to me.

PR #323, opened 2017-01-15
Help doc for Event View

dacharyc:
This is a markdown file that contains a Help doc for the Event view,
plus images to support the Help doc.

This requires Kay’s changes to the Help Docs controller.

PR #325, opened 2017-01-18
fix Event.upcoming

pcai:
Unfortunately the logic for `Event#upcoming` appears to have been reversed. I believe this change was introduced in https://github.com/ReadyResponder/ReadyResponder/pull/312 

Behavior in development branch: `Event.upcoming` returns events that have already ended
Behavior in this PR: `Event.upcoming` returns events that have not yet ended

This also fixes the "unrelated" notification spec failure in https://github.com/ReadyResponder/ReadyResponder/pull/315 , https://github.com/ReadyResponder/ReadyResponder/pull/323 and https://github.com/ReadyResponder/ReadyResponder/pull/324 because the "new notification" form populates dropdowns with "upcoming" events.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -28,7 +28,7 @@ class Event < ActiveRecord::Base
   accepts_nested_attributes_for :timecards
   accepts_nested_attributes_for :certs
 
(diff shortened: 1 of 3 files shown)


kgf (commented):
Looks good, thanks for the additional test

PR #326, opened 2017-01-20
Lots of small UI tweaks to make the UI consistient.

kgf:
This included removing the 'New" links for the top menu, and making all
actions be in the sidebar. It also included some tuning of which columns
are displayed in the datatables


app/controllers/people_controller.rb
```diff
@@ -35,7 +35,7 @@ def everybody
   end
 
   def other
-    @people = Person.active.select{|person| person.department&.name != 'CERT' && person.department&.name != 'Police'}
+    @people = Person.active.joins(:department).where("departments.manage_people": false)
```
pcai, line 38: oh nice change

spec/features/notifications_spec.rb
```diff
@@ -6,7 +6,7 @@
 
     context "from an event" do
       let!(:event)  { create(:event) }
-      it "should create a notification for that event" do
+      pending "should create a notification for that event" do
```
pcai, line 9: it may not look like it, but im fairly certain this is fixed by https://github.com/ReadyResponder/ReadyResponder/pull/325 - I can cherry-pick this to double check.

kgf: I could NOT get that sucker to pass. The controller could not find the parent event, even though the spec was setting it up.

kgf:
Whoever gets to it first, please merge. I needed to make sure this got in first, so that I could train to this UI. I will now be looking to reduce the backlog of PRs.

pcai:
I just checked out this branch and clicked around in the ui. code looks fine and no obvious defects introduced. But I did merge in https://github.com/ReadyResponder/ReadyResponder/pull/325 and that troublesome notifications spec now passes:

```
$ rspec spec/features/notifications_spec.rb
/home/pcai/lims/config/initializers/constants.rb:2: warning: already initialized constant GOOGLE_MAPS_API_KEY
/home/pcai/lims/config/environments/test.rb:47: warning: previous definition of GOOGLE_MAPS_API_KEY was here

Randomized with seed 2987

Notifications
  visit notifications
    from an event
      should create a notification for that event

Top 1 slowest examples (1.03 seconds, 88.2% of total time):
  Notifications  visit notifications from an event should create a notification for that event
    1.03 seconds ./spec/features/notifications_spec.rb:9

Finished in 1.17 seconds (files took 2.16 seconds to load)
1 example, 0 failures

Randomized with seed 2987
```

I think that means the failure is legitimate - I tried this from the ui and the notification really isn't being saved, because if there are no events that match the scope Event.upcoming, that select dropdown will never be populated with a valid event.

Issue #321, opened 2017-01-15
When looking at the target of the Events link, too many are shown


kgf:
By default, it should only show events that are scheduled or in-session. Getting old events should be an additional action. The additional action should appear as a blue button on the left side navigation..

sergeimellow:
I just took a look at this, does this solution suffice? 
https://github.com/ReadyResponder/ReadyResponder/pull/328/files

kgf:
If you could just resolve the issue with having the same page title, that would work.
Thanks

kgf:
Resolved by #328

PR #328, opened 2017-01-20
Make current events default

sergeimellow:
add current events as default to events index view and add buttons to toggle between current and all events
#### All Events
image
#### Current Events
image

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -15,8 +15,9 @@ class EventsController < ApplicationController
   # GET /events
   # GET /events.json
   def index
-    @events = Event.all
-    @page_title = "All Events"
+    # by default show all scheduled or in-session events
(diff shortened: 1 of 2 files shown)


app/controllers/events_controller.rb
```diff
@@ -15,8 +15,9 @@ def schedule
   # GET /events
   # GET /events.json
   def index
-    @events = Event.all
-    @page_title = "All Events"
+    # by default show all scheduled or in-session events
+    @events = params['all_events'] == "true" ? Event.all : Event.where('end_time > ?', Time.now)
+    @page_title = "Scheduled and In-Session Events"
```
kgf: I don't think this will work with a ternary. Won't this code give the same page title for both conditions ? That is, when I click "All Events" the page title will still be "Scheduled"

sergeimellow: good catch, resolved.

kgf:
Resolved #321

Issue #336, opened 2017-01-26
Need to remove phone number hack


kgf:
This was put in to match the data in a production system. That data has now been corrected, and this can be removed. 

Validation should be added to ensure phone numbers are entered correctly.

https://github.com/ReadyResponder/ReadyResponder/blob/development/app/models/person.rb#L107

kgf:
Closed by #338

PR #338, opened 2017-01-27
Phone numbers now need to adhere to the ITU standards

kgf:
In the US, this means "+18005551212" style.


masukomi:
so, without a validation to provide errors when people _don't_ enter numbers that adhere i feel this is just going to be a frustrating issue for users.

masukomi:
also, not all countries have phone numbers of the same length and so the `2..12` is going to be an issue, and I don't see how you're going to distinguish between `+1` country code  followed by a `2` and the `+12` country code (and similar situations).

kgf:
I'll get it working in the US first. Agreed regarding needing a validation

kgf:
Addresses #336

kgf:
There was weirdness in this PR. It has been replaced by PR #356

PR #370, opened 2017-02-26
adds description to item_type, resolves #345

### LINKED ISSUE #345: Item Type needs description field
Item Type needs description field

abrophy:


kgf:
Awesome, thanks for your help !

kgf:
@abrophy 
I reverted this, as I merged it too quickly, and there was no migration to change the schema. This would end up being a breaking change for folks. 
Could you please add it and re-submit the PR. 
Thanks !
Kevin

abrophy:
I didn't change the schema as there is already a string field in the item_types table for "description", if that's what you mean. If not what other changes need to be made to the schema?

kgf:
It was my mistake, there was an earlier migration to add the column to the table. I reverted the reversion.

PR #374, opened 2017-02-26
Event filters availabilities from inactive users, fixes #359

### LINKED ISSUE #359: Person status does not update Availability
If an availability has been entered for a person for an event, and the person's status changes to inactive, they continue to appear as available for the event.

Their availability should not be returned at all if they are inactive.

Code wise: needs to be scoped to active people
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/models/event.rb#L71

abrophy:
resolves #359


kgf:
I actually prefer the way you did it. `for_time_span` should not, IMO, have anything about people being active. That would seem to be too tightly coupled.

The scope is nice.

PR #379, opened 2017-03-08
Remove extra constant assignment for Google Maps

mxie:


pcai:
this is safe to remove because its also declared in `initializers/constants.rb` right?

kgf:
I believe that's true, Peter.

PR #381, opened 2017-03-08
Omit title object from new title route path

### LINKED ISSUE #316: Bad URL when creating new title after creating new title
I created a new title, was taken back to the title page, and clicked the new title button again...

This time, there's a .3 in the URL that shouldn't be there.

screen shot 2017-01-15 at 11 00 25 am

mxie:
Because the route for `new_title_path` doesn't expect an id to be a part
of it, it's treating the passed object as an indicator of the view
rendering format (e.g. html, json, etc.). We also just don't need the
object at all.

Fixes #316.


pcai (approved):

pcai:
looks good to me

PR #383, opened 2017-03-10
Availability sets older duplicates to Cancelled, fixes #372

### LINKED ISSUE #372: Availabilities are not unique for one person
Currently, as of v2.2, someone can send in a text message and create an availability. This works well.

But what happens when they send in another ? It simply adds it without considering what is already in the system. For example, let's say Matt Udy sends in that he's available for Monday, noon to five. He later finds out he has to work, so he sends in `Unavailable` for the same time period. The system will count him as both available ***and*** unavailable. 

As with most things related to time, there are a lot of edge cases, but I'd like to keep this simple at first.

If someone sends in a second Availability record, we should set the status of the first one to `Cancelled` and then record the second one.

The edge cases will come when the times do not line up exactly.

abrophy:
This method naively compares start and end times.

I can have it round to the closest 1/4 hour but I wasn't sure if we have an abstraction for rounding of time we're using in the project.

Can easily implement that in the method, or just do the rounding in the method itself, whatever you prefer.

### DIFF
--- a/app/models/availability.rb
+++ b/app/models/availability.rb
@@ -48,7 +48,7 @@ class Availability < ActiveRecord::Base
     previous_availabilities = person.availabilities.for_time_span(start_time..end_time).order(:created_at)
     previous_availabilities.each do |a|
       if start_time == a.start_time && end_time == a.end_time
-        a.update(status: "Cancelled")
+        a.update(status: "Cancelled") unless self == a
       end
     end
   end
--- a/app/models/availability.rb
+++ b/app/models/availability.rb
@@ -1,7 +1,7 @@
 class Availability < ActiveRecord::Base
   belongs_to :person
 
-  before_create :cancel_duplicates
+  after_create :cancel_duplicates
 
   validates_presence_of :person, :status, :start_time, :end_time
   validates_chronology :start_time, :end_time
--- a/app/models/availability.rb
+++ b/app/models/availability.rb
@@ -1,6 +1,8 @@
 class Availability < ActiveRecord::Base
   belongs_to :person
 
+  before_create :cancel_duplicates
+
   validates_presence_of :person, :status, :start_time, :end_time
   validates_chronology :start_time, :end_time
 
@@ -38,5 +40,16 @@ class Availability < ActiveRecord::Base
         count = Availability.where('date(start_time) <= ? AND date(end_time) >= ?', date, date).count
         { 'Date' => date, 'Count' => count }
       end
+  end
+
+  private
+
+  def cancel_duplicates
+    previous_availabilities = person.availabilities.for_time_span(start_time..end_time).order(:created_at)
+    previous_availabilities.each do |a|
+      if start_time == a.start_time && end_time == a.end_time
+        a.update(status: "Cancelled")
+      end
+    end
   end
 end
--- a/spec/models/availability_spec.rb
+++ b/spec/models/availability_spec.rb
@@ -34,4 +34,16 @@ RSpec.describe Availability, type: :model do
       expect(availability.partially_available?(event)).to eq(false)
     end
   end
+
+  context 'cancel_duplicates' do
+    it 'sets status for previous availabilities with matching start and end times to Cancelled' do
+      start_time = Time.now
+      end_time = start_time + 2.minutes
+      first_availability = create(:availability, person: a_person, start_time: start_time, end_time: end_time, created_at: Time.now - 2.minutes)
+      second_availability = create(:availability, person: a_person, start_time: start_time, end_time: end_time)
+
+      expect(first_availability.reload.status).to eq('Cancelled')
+      expect(second_availability.status).to eq('Unavailable')
+    end
+  end
 end
--- a/spec/models/availability_spec.rb
+++ b/spec/models/availability_spec.rb
@@ -39,10 +39,11 @@ RSpec.describe Availability, type: :model do
     it 'sets status for previous availabilities with matching start and end times to Cancelled' do
       start_time = Time.now
       end_time = start_time + 2.minutes
-      first_availability = create(:availability, person: a_person, start_time: start_time, end_time: end_time, created_at: Time.now - 2.minutes)
+      first_availability = create(:availability, person: a_person, start_time: start_time, end_time: end_time)
       second_availability = create(:availability, person: a_person, start_time: start_time, end_time: end_time)
 
-      expect(first_availability.reload.status).to eq('Cancelled')
+      first_availability = first_availability.reload
+      expect(first_availability.status).to eq('Cancelled')
       expect(second_availability.status).to eq('Unavailable')
     end
   end


kgf:
I can't look right now, but I believe we do have code for rounding to the nearest quarter hour.

Sadly, the tests did not pass.

abrophy:
I'll get on this right away.

app/models/availability.rb
```diff
@@ -1,6 +1,8 @@
 class Availability < ActiveRecord::Base
   belongs_to :person
 
+  before_create :cancel_duplicates
```
kgf: I think this would be better as an after save hook. We would want it on save, so that if someone updated an availability to overlap, we'd still catch that. And an after hook, so that if the availability _doesn't_ save for some reason, we won't have cancelled availabilities we should not have.

spec/models/availability_spec.rb
```diff
@@ -34,4 +34,16 @@
       expect(availability.partially_available?(event)).to eq(false)
     end
   end
+
+  context 'cancel_duplicates' do
+    it 'sets status for previous availabilities with matching start and end times to Cancelled' do
+      start_time = Time.now
+      end_time = start_time + 2.minutes
+      first_availability = create(:availability, person: a_person, start_time: start_time, end_time: end_time, created_at: Time.now - 2.minutes)
```
kgf: Sorry, I'm not following this. Why are you setting the created_at ?

spec/models/availability_spec.rb
```diff
@@ -34,4 +34,16 @@
...
+      second_availability = create(:availability, person: a_person, start_time: start_time, end_time: end_time)
+
+      expect(first_availability.reload.status).to eq('Cancelled')
```
kgf: Needing the reload here makes sense to me. Line 42 saves a record into the database, and loads a copy into memory, referenced by _first_availability_. Your hook then changes the underlying persisted record, but does not reload the object.
You could `first_availability.reload`, then `expect(first_availability.status).to eq("Cancelled")`

spec/models/availability_spec.rb
abrophy: Is unnecessary, you're right.

spec/models/availability_spec.rb
abrophy: I'll give this a try

app/models/availability.rb
abrophy: I agree with this, will do.

spec/models/availability_spec.rb
```diff
@@ -39,10 +39,11 @@
     it 'sets status for previous availabilities with matching start and end times to Cancelled' do
       start_time = Time.now
       end_time = start_time + 2.minutes
-      first_availability = create(:availability, person: a_person, start_time: start_time, end_time: end_time, created_at: Time.now - 2.minutes)
+      first_availability = create(:availability, person: a_person, start_time: start_time, end_time: end_time)
       second_availability = create(:availability, person: a_person, start_time: start_time, end_time: end_time)
 
-      expect(first_availability.reload.status).to eq('Cancelled')
+      first_availability = first_availability.reload
```
abrophy: moving the reload into its own expression like this still hasn't fixed the issue.
Test is read on Travis but green on local machine.

app/models/availability.rb
```diff
@@ -39,4 +41,15 @@ def self.process_data
...
       end
   end
+
+  private
+
+  def cancel_duplicates
+    previous_availabilities = person.availabilities.for_time_span(start_time..end_time).order(:created_at)
+    previous_availabilities.each do |a|
+      if start_time == a.start_time && end_time == a.end_time
+        a.update(status: "Cancelled") unless self == a
```
pcai, line 51: Is there a reason you are using `for_time_span(start_time..end_time)` instead of just using an equality search to get precisely the records that need to be cancelled? e.g., something like `person.availabilities.where(start_time: self.start_time, end_time: self.end_time).where.not(id: self.id)`

I think that way you can omit `unless self == a` and the time comparison, since you'll avoid loading the `self` record and anything that doesnt match.

kgf: This would be a fine first approach. Right now folks have no way to correct a mistake without involving someone who has access to the web UI. The most common use case is an availability for a specific event, so the times would exactly match.

pcai:
The tests are probably failing on travis for one of 2 reasons:
* the time resolution is not correct. saving and retrieving from the database might cause the equality comparison to fail.
* a timezone issue is causing the save and retrieval to make the equality comparison fail.

I'm not familiar enough with travis to know if there's an ssh/debug option but you could always add a debug commit that ups the verbosity to confirm or deny

app/models/availability.rb
pcai: Right. If that's the case, my suggestion is to replace this whole block with something like: 
```ruby
# cancel all availabilities that are exact duplicates of this one
person.availabilities.where(start_time: self.start_time, end_time: self.end_time)
  .where.not(id: self.id).update_all(status: "Cancelled")
```

I haven't tested this but I think that captures the essence of what this is trying to accomplish?

pcai: Sorry, just saw the linked issue suggests the other approach. I'll take a look to see if I can shed any light on the travis failure

spec/models/availability_spec.rb
```diff
@@ -34,4 +34,17 @@
       expect(availability.partially_available?(event)).to eq(false)
     end
   end
+
+  context 'cancel_duplicates' do
+    it 'sets status for previous availabilities with matching start and end times to Cancelled' do
```
mxie, line 39: Would you mind rearranging the phrasing of this to say "sets status to X for..."? It took me a few reads before realizing what it was saying. 😬

spec/models/availability_spec.rb
```diff
@@ -34,4 +34,17 @@
...
+      start_time = Time.now
+      end_time = start_time + 2.minutes
+      first_availability = create(:availability, person: a_person, start_time: start_time, end_time: end_time)
+      second_availability = create(:availability, person: a_person, start_time: start_time, end_time: end_time)
+
+      first_availability = first_availability.reload
```
mxie, line 45: Do we have to reload?

spec/models/availability_spec.rb
```diff
@@ -34,4 +34,17 @@
       expect(availability.partially_available?(event)).to eq(false)
     end
   end
+
+  context 'cancel_duplicates' do
+    it 'sets status for previous availabilities with matching start and end times to Cancelled' do
+      start_time = Time.now
```
mxie, line 40: We should use `Time.current` instead: https://robots.thoughtbot.com/its-about-time-zones

spec/models/availability_spec.rb
mxie: Nevermind - I see why now. :)

app/models/availability.rb
mxie: I agree with what @pcai has said about rewriting the block. If we want to use `for_time_span`, I think it'd look something like this:

```ruby
person.availabilities.for_time_span(start_time..end_time).where.not(id: id).update_all(status: "Cancelled")
```

kgf:
@abrophy I was able to get this to reliably pass by making these changes.
https://github.com/ReadyResponder/ReadyResponder/pull/399/files/8b9f45c76e6c6ff8c198e2534d0103d56fd86e4a..e7ab945191e710092de1f3fa35b3e78d160b0e05
I doubt the removal of the `order` clause was important, but it was there so....
Would you like to make the changes and update this PR, or perhaps you're busy and would prefer that I merge #399 .
I'd like to cut a release soon, so just let me know. 
Thanks for your help !
Kevin

abrophy:
Sorry I had other projects come up in the mean time,  thanks to everyone who looked over this

PR #385, opened 2017-03-20
Timecards managed via text message

### LINKED ISSUE #380: Timecards managed via text message
Currently, timecards can be entered via the web UI. This creates the need for someone to do the clerical data entry. It would be much better if each individual volunteer could enter their own timecard through a text message.

People would text `arrive` to the ReadyResponder number when they arrive at the staging area. The capitalization should not matter. Upon receiving this message, the system should look up who sent the message, and see if there are any existing incomplete timecards for that person. These should be set to an `error` status. If no other timecards are found, create a new timecard for that person. The timecard should have the start time based on the time of the text. The status should be set to Incomplete. The system should respond to the user that a timecard has been started.

People would text `depart` when they are done and leaving the staging area. The system should look up the person who sent the message. It should then look for an open timecard, one that is `incomplete` and only has a start time. If more or less than one is found, they should all be set to an `error` status. The response should include an error message. Even if there is no correct timecard to complete, a new timecard with only this end_time should still be created, and immediately set to an error status. If only one is found, that timecard should be updated with the end time from the message, and a confirmation message returned. 

ngrok is helpful in developing these features.

pcai:
resolves https://github.com/ReadyResponder/ReadyResponder/issues/380.

Tweaks from the specced behavior: 
> Upon receiving this message, the system should look up who sent the message, and see if there are any existing incomplete timecards for that person. These should be set to an error status. If no other timecards are found, create a new timecard for that person.

The new behavior, as discussed, is to always create a new timecard, but before doing so, it will set all existing incomplete timecards to "error" first.

(previously, my tweak would cause it to no-op, which would make it idempotent, but also impossible to start a new timecard if they forgot to sign out from a previous shift).


app/services/message/extract_keyword.rb
```diff
@@ -17,15 +17,14 @@ def call
...
     # TODO Additional keywords => maybe rcv received ack
     # TODO tasks task roster (who's assigned to event)
     # TODO schedule (What I'm scheduled for)
     # TODO onduty on-duty for who's on duty right now. & signin signout
     # TODO "crew id_code" to allow supervisor to see who has responded.
-    if white_list.keys.include? first_word
-      return white_list[first_word]
-    end
-    return nil
+    return white_list[first_word]
```
pcai, line 28: Not strictly necessary for this pr, but I think this is more concise and perhaps faster. Tests continue to pass.

kgf:
Peter, thanks for your help. 
I gave this some thought. Disasters are, by definition, chaotic. People will sign-in, work for quite a while, forget to sign out when they leave. Then they come back, they'll sign in again. We should not throw that away. Therefore, I think the best solution is:

- When someone is trying to sign-in
- and there is already one or more incomplete timecards
- all their incomplete timecards should be set to error status
- and a new incomplete timecard created

After the event has calmed down, someone can review timecards in error and resolve them manually. The important thing is that the resolution can be done asynchronously.

Does that make sense ?
Kevin

pcai:
OK. That explanation makes sense to me, I'll make the change -- thanks

pcai:
Updated with requested change to the 'arrive' behavior.

app/services/msg/depart.rb
```diff
@@ -0,0 +1,28 @@
...
+        # set all incomplete timecards to error state
+        incomplete_timecards.update_all(status: "Error")
+        # record the depart time as a new timecard entry
+        Timecard.create!(person_id: @person.id,
+                         status: "Error",
+                         end_time: Time.now)
+        return "Error: #{timecards_count} incomplete timecards."
+      end
+      
+      timecard = incomplete_timecards.first
```
kgf: On my first scan, I missed that we would not fall through to line 22, due to line 19. What do you think of doing the happy path ( timecards_count =1 ) as an if block, and then an else block for not equaling one ? Lines 13-19 would be in the else block. I realize your code has the same effect, but I think that would better illuminate the two paths.

Your thoughts ?
Thanks for your help. I'll be doing a deployment before Wed evening, so having this is great !
Kevin

kgf (commented):
Thanks for your help.

app/services/msg/depart.rb
pcai: makes sense. good suggestion

app/services/message/extract_keyword.rb
kgf: I like it, it certainly reads better

kgf:
Thanks for your help. This will be awesome !

PR #391, opened 2017-04-14
Code Climate config

larkinscott:
Code Climate configuration file with Ruby duplication and Rubocop enabled.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,15 @@
+engines:
+ duplication:
+  enabled: true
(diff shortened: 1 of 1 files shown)


kgf:
Thanks for your help !

kgf:
@larkinscott I have not yet seen this appear as a new check on incoming PRs. Is there another step I need to take ?

kgf:
@larkinscott still having issue where code climate does not appear as one of the checks.

PR #393, opened 2017-04-27
Removed departments from item categories.

kgf:
Vehicles are vehicles, no matter who owns them. Departmental ownership is on the individual item. The old approach was too cumbersome.

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -9,6 +9,7 @@ class ItemsController < ApplicationController
 
   def show
     @item = Item.find(params[:id])
+    @item_type = @item.item_type
     @last_editor = last_editor(@item)
     @page_title = "Item: #{@item.name}"
 
@@ -17,13 +18,18 @@ class ItemsController < ApplicationController
   end
 
   def new
-    @item = Item.new(status: 'Available')
+    @item_type = ItemType.find ( params[:item_type_id])
+    @item = Item.new(
+            status: 'Available',
+            item_type_id: params[:item_type_id])
+
     @page_title = "New Item"
   end
 
   def edit
     @item = Item.find(params[:id])
-    @page_title = "Editting Item #{@item.name}"
+    @item_type = @item.item_type
+    @page_title = "Editing Item #{@item.name}"
   end
 
   def create
--- a/app/models/item.rb
+++ b/app/models/item.rb
@@ -37,4 +37,9 @@ class Item < ActiveRecord::Base
     repairs.where("service_date > ?", 6.months.ago).pluck(:cost).compact.inject(:+) || 0
   end
 
+  def location_name
+    location.name if location
+    "Unknown"
+  end
+  
 end
--- a/app/views/items/edit.html.erb
+++ b/app/views/items/edit.html.erb
@@ -1,6 +1,6 @@
 <% content_for :sidebar do %>
   <h1>Editing Item</h1>
-  <%= sidebar_button_link 'Cancel Edit', :back %>
+  <%= sidebar_button_link 'Cancel', item_type_path(@item_type) %>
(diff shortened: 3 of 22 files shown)


app/controllers/items_controller.rb
```diff
@@ -17,13 +18,18 @@ def show
   end
 
   def new
-    @item = Item.new(status: 'Available')
+    @item_type = ItemType.find ( params[:item_type_id])
```
masukomi, line 21: I don't think there's anything really "wrong" with this method but it doesn't feel right to me. 

It's weird that you get the `@item_type` as a stand-alone thing even though it's theoretically part of the `@item` that we're working with.

Similarly it feels odd that it's being used as it is in the view layer. I think I would rather see:

```

item_type = ItemType.find( params[:item_type_id])
@item = Item.new(status: 'Available',
                               item_type: item_type)
```

app/controllers/items_controller.rb
```diff
@@ -17,13 +18,18 @@ def show
...
+    @item = Item.new(
+            status: 'Available',
+            item_type_id: params[:item_type_id])
+
     @page_title = "New Item"
```
masukomi, line 26: I don't really understand this. The controller shouldn't really be concerned with non-dynamic presentation stuff like this. I'd just hardcode it in the view like all the other text things until you've got translations, and then i'd put the call to the translation in the view too, not the controller.

app/models/item.rb
```diff
@@ -37,4 +37,9 @@ def recent_costs
     repairs.where("service_date > ?", 6.months.ago).pluck(:cost).compact.inject(:+) || 0
   end
 
+  def location_name
+    location.name if location
+    "Unknown"
```
masukomi, line 42: Doesn't this ALWAYS return "Unknown" ?!

did you mean `return location.name if location` on the line above?

app/controllers/items_controller.rb
```diff
@@ -17,13 +18,18 @@ def show
...
   end
 
   def edit
     @item = Item.find(params[:id])
```
masukomi, line 30: add the `.includes(:item_type)` to save the 2nd query happening on the next line.

app/controllers/items_controller.rb
```diff
@@ -17,13 +18,18 @@ def show
...
-    @page_title = "Editting Item #{@item.name}"
+    @item_type = @item.item_type
```
masukomi, line 31: again, This feels wrong to me. This page is about items and the item has the ItemType in it so, i'd just ask the item for its item type in the view when needed instead of making a second var for it. 

Also, removing these unnecessary 2nd vars means you have less things to test the presence of in a controller  / integration test.

app/views/items/edit.html.erb
```diff
@@ -1,6 +1,6 @@
 <% content_for :sidebar do %>
   <h1>Editing Item</h1>
-  <%= sidebar_button_link 'Cancel Edit', :back %>
+  <%= sidebar_button_link 'Cancel', item_type_path(@item_type) %>
```
masukomi, line 3: i'd prefer `item_type_path(@item.item_type)` (and getting rid of `@item_type` in the controller.

app/views/items/new.html.erb
```diff
@@ -1,6 +1,6 @@
 <% content_for :sidebar do %>
   <h1>New Item</h1>
-  <%= sidebar_button_link 'Cancel', items_path %>
+  <%= sidebar_button_link 'Cancel', item_type_path( @item_type) %>
```
masukomi, line 3: `item_type_path(@item.item_type)`

masukomi (changes requested):

kgf:
Superseded by #394

PR #394, opened 2017-05-02
Add condition to item and ui inprovements

kgf:
This supersedes PR #393, since I accidentally made the changes requested by @masukomi  in the wrong branch.

### DIFF
--- a/app/controllers/item_categories_controller.rb
+++ b/app/controllers/item_categories_controller.rb
@@ -16,8 +16,7 @@ class ItemCategoriesController < ApplicationController
 
   # GET /item_categories/new
   def new
-    @item_category = ItemCategory.new
-    @item_category.status = 'Active'
+    @item_category = ItemCategory.new(status: 'Active')
   end
 
   # GET /item_categories/1/edit
--- a/app/controllers/item_categories_controller.rb
+++ b/app/controllers/item_categories_controller.rb
@@ -1,12 +1,12 @@
 class ItemCategoriesController < ApplicationController
   before_filter :authenticate_user!
   load_and_authorize_resource
-  
+
   before_action :set_item_category, only: [:show, :edit, :update, :destroy]
 
   # GET /item_categories
   def index
-    @item_categories = ItemCategory.all
+    @item_categories = ItemCategory.all.order(:name)
   end
 
   # GET /item_categories/1
@@ -17,6 +17,7 @@ class ItemCategoriesController < ApplicationController
   # GET /item_categories/new
   def new
     @item_category = ItemCategory.new
+    @item_category.status = 'Active'
   end
 
   # GET /item_categories/1/edit
--- a/app/controllers/repairs_controller.rb
+++ b/app/controllers/repairs_controller.rb
@@ -22,12 +22,10 @@ class RepairsController < ApplicationController
   def create
     @repair = Repair.new(params[:repair])
 
-    respond_to do |format|
-      if @repair.save
-        redirect_to @repair, notice: 'Repair was successfully created.'
-      else
-        render action: "new"
-      end
+    if @repair.save
+      redirect_to @repair, notice: 'Repair was successfully created.'
+    else
+      render action: "new"
     end
   end
 
--- a/app/controllers/repairs_controller.rb
+++ b/app/controllers/repairs_controller.rb
@@ -7,9 +7,8 @@ class RepairsController < ApplicationController
   end
 
   def show
-    @item = @repair.item
-    @item_type = @item.item_type
-    @item_category = @item_type.item_category
+    @last_editor = last_editor(@repair)
+    @page_title = "Repair for #{@repair.item}"
   end
 
   def new
--- a/app/controllers/repairs_controller.rb
+++ b/app/controllers/repairs_controller.rb
@@ -2,86 +2,47 @@ class RepairsController < ApplicationController
   before_filter :authenticate_user!
   load_and_authorize_resource
 
-  # GET /repairs
-  # GET /repairs.json
   def index
     @repairs = Repair.all
-
-    respond_to do |format|
-      format.html # index.html.erb
-      format.json { render json: @repairs }
-    end
   end
 
-  # GET /repairs/1
-  # GET /repairs/1.json
   def show
-    @repair = Repair.find(params[:id])
-
-    respond_to do |format|
-      format.html # show.html.erb
-      format.json { render json: @repair }
-    end
+    @item = @repair.item
+    @item_type = @item.item_type
+    @item_category = @item_type.item_category
   end
 
-  # GET /repairs/new
-  # GET /repairs/new.json
   def new
     @repair = Repair.new
     @repair.item_id = params[:item_id].presence
-
-    respond_to do |format|
-      format.html # new.html.erb
-      format.json { render json: @repair }
-    end
   end
 
-  # GET /repairs/1/edit
   def edit
-    @repair = Repair.find(params[:id])
   end
 
-  # POST /repairs
-  # POST /repairs.json
   def create
     @repair = Repair.new(params[:repair])
 
     respond_to do |format|
       if @repair.save
-        format.html { redirect_to @repair, notice: 'Repair was successfully created.' }
-        format.json { render json: @repair, status: :created, location: @repair }
+        redirect_to @repair, notice: 'Repair was successfully created.'
       else
-        format.html { render action: "new" }
-        format.json { render json: @repair.errors, status: :unprocessable_entity }
+        render action: "new"
       end
     end
   end
 
-  # PUT /repairs/1
-  # PUT /repairs/1.json
   def update
-    @repair = Repair.find(params[:id])
-
     respond_to do |format|
       if @repair.update_attributes(params[:repair])
-        format.html { redirect_to @repair.item, notice: 'Repair was successfully updated.' }
-        format.json { head :no_content }
+        redirect_to @repair.item, notice: 'Repair was successfully updated.'
       else
-        format.html { render action: "edit" }
-        format.json { render json: @repair.errors, status: :unprocessable_entity }
+        render action: "edit"
       end
     end
   end
 
-  # DELETE /repairs/1
-  # DELETE /repairs/1.json
   def destroy
-    @repair = Repair.find(params[:id])
     @repair.destroy
-
-    respond_to do |format|
-      format.html { redirect_to repairs_url }
-      format.json { head :no_content }
-    end
   end
 end
--- a/app/helpers/items_helper.rb
+++ b/app/helpers/items_helper.rb
@@ -10,7 +10,7 @@ module ItemsHelper
 
   def edit_item_button
     return nil if cannot? :edit, @item
-    return link_to "Edit #{@item.name}",
+    return link_to "Edit #{@item}",
                    edit_item_path(@item),
                    class: 'btn btn-block btn-primary'
   end
--- a/app/helpers/items_helper.rb
+++ b/app/helpers/items_helper.rb
@@ -1,6 +1,11 @@
 module ItemsHelper
+
+  def item_condition_label(item)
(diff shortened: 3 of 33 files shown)


app/controllers/item_categories_controller.rb
```diff
@@ -17,6 +17,7 @@ def show
   # GET /item_categories/new
   def new
     @item_category = ItemCategory.new
+    @item_category.status = 'Active'
```
masukomi: I think `@item_category = ItemCategory.new(status: 'Active')` would have been a better way to do this.

app/controllers/repairs_controller.rb
```diff
@@ -2,86 +2,46 @@ class RepairsController < ApplicationController
...
-  # GET /repairs/1
-  # GET /repairs/1.json
   def show
-    @repair = Repair.find(params[:id])
-
-    respond_to do |format|
-      format.html # show.html.erb
-      format.json { render json: @repair }
-    end
+    @last_editor = last_editor(@repair)
```
masukomi, line 10: side note. i find it extremely weird / wrong that `last_editor` doesn't return a user. Violates rule of least surprise and totally limits its utility. Now, maybe a `last_editor` method and a `last_editor_name` method would be a good combo, the 2nd building on the output of the 1st.

obviously that's outside of the scope of this PR but since it was being used here i went and investigated.

app/controllers/repairs_controller.rb
```diff
@@ -2,86 +2,46 @@ class RepairsController < ApplicationController
...
+    @page_title = "Repair for #{@repair.item}"
```
masukomi, line 11: because you don't have a custom `to_s` on `Item` I'm surprised that this isn't `@repair.item.name`

app/controllers/repairs_controller.rb
```diff
@@ -2,86 +2,46 @@ class RepairsController < ApplicationController
...
 
     respond_to do |format|
       if @repair.save
-        format.html { redirect_to @repair, notice: 'Repair was successfully created.' }
-        format.json { render json: @repair, status: :created, location: @repair }
+        redirect_to @repair, notice: 'Repair was successfully created.'
       else
-        format.html { render action: "new" }
-        format.json { render json: @repair.errors, status: :unprocessable_entity }
+        render action: "new"
```
masukomi: if the repair failed to save shouldn't there be an error message saying why? No rails flash?

app/controllers/repairs_controller.rb
```diff
@@ -2,86 +2,46 @@ class RepairsController < ApplicationController
...
-
     respond_to do |format|
       if @repair.update_attributes(params[:repair])
-        format.html { redirect_to @repair.item, notice: 'Repair was successfully updated.' }
-        format.json { head :no_content }
+        redirect_to @repair.item, notice: 'Repair was successfully updated.'
       else
-        format.html { render action: "edit" }
-        format.json { render json: @repair.errors, status: :unprocessable_entity }
+        render action: "edit"
```
masukomi, line 37: again, why no message to user as to why update failed?

app/helpers/items_helper.rb
```diff
@@ -1,11 +1,16 @@
...
+
   def item_status_label(item)
-    content_tag(:span, item.status, class: item_label_class(item))
+    content_tag(:span, item.status, class: item_status_class(item))
   end
 
   def edit_item_button
     return nil if cannot? :edit, @item
-    return link_to 'Edit Item',
+    return link_to "Edit #{@item.name}",
```
masukomi: here you used `@item.name` but above you used `.item` i think this reinforces that the other one should be changed.

app/models/repair.rb
```diff
@@ -4,9 +4,14 @@ class Repair < ActiveRecord::Base
                   :status, :user_id, :cost
   belongs_to :item
   belongs_to :person
-  validates_numericality_of :cost
+  delegate :item_category, :to => :item
+  delegate :item_type, :to => :item
```
masukomi, line 8: i ... don't really understand why these exist on the repair even if they're delegated. I don't see why the repair cares what "type" of item. 

For example. if the "repair" is fixing a wheel it doesn't matter so much what _kind_ of car it is that gets repaired. it matters _which_ car it is.

app/views/items/index.html.erb
```diff
@@ -18,10 +19,10 @@
   <tbody>
     <% @items.each do |item| %>
       <tr>
-        <td><%= item.location.name if item.location %></td>
         <td><%= link_to item.name, item_path(item) %></td>
         <td><%= "#{item.brand} #{item.model}" %></td>
         <td><%= item.category %></td>
+        <td><%= item.location_name %></td>
```
masukomi, line 25: this still feels a little uncomfortable to me but i don't currently have a better suggestion. 

It's not just that it knows about an internal thing of location it's just the concept that an item has any clue about the NAME of the place it's located in. Know where it is, sure. Know the name of the place where it is? ... weird.

app/views/items/show.html.erb
```diff
@@ -1,17 +1,27 @@
...
-  <%= sidebar_button_link 'Return to Listing', items_path %>
   <%= edit_item_button %>
   <%= add_repair_button %>
   <%= add_inspection_button %>
 <% end %>
 
 <h3>
+  <%= link_to @item.item_category,
+                item_category_path(@item.item_category) %> >
+  <%= link_to @item.item_type, @item.item_type %> >
```
masukomi: this just looks like a typo.  It would look less like a typo if it was

    link_to @item.item_type.name, @item.item_type

and even better if it was 

    link_to @item_type.name, item_type_path(@item.item_type)

This would also be more consistent with how you're making links elsewhere.

app/views/items/show.html.erb
```diff
@@ -1,17 +1,27 @@
...
+</h3>
+
+<h3>
   <% if @item.status.present? %>
     <%= item_status_label(@item) %> &nbsp;
   <% end %>
-  <%= @item.name %>
+
+  <% if @item.condition.present? %>
+    <%= item_condition_label(@item) %> &nbsp;
```
masukomi, line 20: better to use css padding than `&nbsp;` 

    <span class="item_condition_label"><%=  item_condition_label(@item) %></span>

and have `item_condition_label` be something like `display: inline-block; padding-right 1rem;`

app/views/items/show.html.erb
```diff
@@ -1,17 +1,27 @@
...
+  <% end %>
+  <br/><br/>
   <% if @item.brand.present? || @item.model.present? %>
-    ( <%=@item.brand.to_s %> <%= @item.model.to_s %> )
+    <%=@item.brand.to_s %> <%= @item.model.to_s %>
```
masukomi: `.to_s` is redundant in these. My understanding is that `<%=` is a fancy way of saying "call `.to_s` on whatever my guts spit out."

spec/features/items_spec.rb
```diff
@@ -1,26 +1,33 @@
...
+    visit edit_item_type_item_path(item_type, @item)
     expect(page).to have_field("item_model", :with => @item.model)
   end
 
   it "a new item form with proper departments filled in" do
-    @department1 = Department.create(name: "Manages items", shortname: "Manages items", manage_items: true)
-    Department.create(name: "Doesn't manage items", shortname: "Doesn't", manage_items: false)
-    visit new_item_path
-    expect(page).to have_select("item_department_id", :options => [@department1.name, ''])
+    department_yes = Department.create(name: "Manages items", shortname: "Manages items", manage_items: true)
```
masukomi, line 26: because you're only using the name, and you're hardcoding the name here there's no need to bother with creating `Department` objects. Instead just replace `department_yes.name` with the name you've hardcoded here. Same for `department_no.name`

This simplifies the test, saves two db writes, and makes things run faster.

spec/models/item_spec.rb
```diff
@@ -5,4 +5,14 @@
     expect(create(:item)).to be_valid
   end
 
+  it 'has a status of OOS if condition is OOS' do
+    item = build :item, condition: 'Out of Service', status: 'Unassigned'
+    expect(item.status).to eq('Out of Service')
```
masukomi: this just seems flat-out wrong. You told it to build you an item with a status of "Unassigned" but then you get a status of "Out of Service" if this passes something is _seriously_ wacky with how this works, and would be very confusing to any developers who aren't you.

spec/models/item_spec.rb
```diff
@@ -5,4 +5,14 @@
...
+  end
+
+  it 'has a status of Unassigned if condition is Available and status is Unassigned' do
+    item = build :item, condition: 'Available', status: 'Unassigned'
+    expect(item.status).to eq('Unassigned')
```
masukomi: this doesn't look wacky, but it does look like you're just testing that FactoryGirl's `build` method did what you asked it too.

the fact that it could be something OTHER than what you told factory girl to build is again VERY surprising (in a bad way) I expect that if i set the status to "foo" the status should stay foo when i query it. Maybe a `before_save` type hook might change it, but there hasn't been anything to trigger something like that. 

I really don't like this behavior.

masukomi (dismissed):

app/views/items/show.html.erb
kgf: I don't have the bandwidth to handle CSS. (Not one of my stronger suits anyway.) CSS will need to be it's own effort.

spec/models/item_spec.rb
kgf: It is testing that an item that is not in a serviceable condition, can not be assigned to a task.

https://github.com/kgf/ReadyResponder/blob/add_condition_to_item_and_ui_inprovements/app/models/item.rb#L39

spec/models/item_spec.rb
kgf: Agreed, something went awry. Will try to figure out where I was going with this.

app/controllers/repairs_controller.rb
kgf: I can appreciate the comment. That could be a good change

app/controllers/repairs_controller.rb
kgf: Added `to_s` on item

app/controllers/repairs_controller.rb
kgf: You get that at no additional cost from `simple_forms`

app/controllers/repairs_controller.rb
kgf: They get a message

app/helpers/items_helper.rb
kgf: Fixed, removed .name here.

app/models/repair.rb
kgf: I need these for the bread crumbs, so as not to have a lot of logic in the view. This is what we discussed in person as well.

app/views/items/index.html.erb
kgf: I think it would be worse to have `item.location.name`

app/views/items/show.html.erb
kgf: I'm dropping the `.name` everywhere, so if I want to change the label, I can do it in one place. 
The form I have is a valid way to specify the target, but I do have it the other way in several places. Also, I am trying to encourage more junior developers, so perhaps the longer version would be better.

Changed

app/views/items/show.html.erb
kgf: Changed

spec/features/items_spec.rb
kgf: The two departments need to be persisted so that the form will pick them up ( or not ) as valid entires in the select boxes. That's what's really being tested.

spec/models/item_spec.rb
kgf: No, actually this is correct. This is testing the happy path, vs the OOS path starting in line 8.

PR #398, opened 2017-05-10
remove 'Template' from CATEGORY_CHOICES in event.rb

itroman:
Fix bug removing 'Template' from category choices

-  Removed 'Template' as an option from the CATEGORY_CHOICES constant in the 
   event model.


kgf:
Thanks for your help !

PR #408, opened 2017-06-26
260 copy tasks when using event template

kgf:
Pulls in the tasks and requirements from a template

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -24,7 +24,7 @@ class EventsController < ApplicationController
 
   def create
     @event = Event.new(event_params)
-    @event.use_template if @event.template.present?
+    @event.use_a_template if @event.template.present?
 
     if @event.save
       redirect_to @event, notice: 'Event was successfully created.'
@@ -43,14 +43,14 @@ class EventsController < ApplicationController
 
   def destroy
     @event.destroy
-    redirect_to events_url
+    redirect_to events_path
   end
 
   private
   def event_params
     params.require(:event).permit(:title, :description, :category,
     :course_id, :duration, :start_time, :end_time, :instructor, :location,
-    :id_code, :status, :timecard_ids, :person_ids, :comments, :is_template, :template_id ,
-    :department_ids => [])
+    :id_code, :status, :timecard_ids, :person_ids, :comments,
+    :is_template, :template_id, department_ids: [])
   end
 end
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -4,13 +4,13 @@ class EventsController < ApplicationController
 
   def index
     # by default show all scheduled or in-session events
-    @events = params['all_events'] == "true" ? Event.all : Event.where('end_time > ?', Time.now)
+    @events = params['all_events'] == "true" ? Event.all : Event.actual.where('end_time > ?', Time.now)
     @page_title = params['all_events'] == "true" ? "All Events" : "Current Events"
   end
 
   def show
-    @event = Event.find(params[:id])
     @page_title = @event.title
+    @last_editor = last_editor(@event)
   end
 
   def new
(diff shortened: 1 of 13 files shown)


app/controllers/events_controller.rb
```diff
@@ -2,97 +2,55 @@ class EventsController < ApplicationController
...
-    respond_to do |format|
-      format.html # index.html.erb
-      format.json { render json: @events }
-    end
   end
 
-  # GET /events/1
-  # GET /events/1.json
   def show
-    @event = Event.find(params[:id])
```
mxie, line 30: Don't we need this?

app/controllers/events_controller.rb
```diff
@@ -2,97 +2,55 @@ class EventsController < ApplicationController
...
+      render action: "edit"
     end
   end
 
-  # DELETE /events/1
-  # DELETE /events/1.json
   def destroy
-    @event = Event.find(params[:id])
     @event.destroy
+    redirect_to events_url
```
mxie: `events_path` perhaps?

app/controllers/events_controller.rb
```diff
@@ -2,97 +2,55 @@ class EventsController < ApplicationController
...
 
-    respond_to do |format|
-      format.html { redirect_to events_url }
-      format.json { head :no_content }
-    end
+  private
+  def event_params
+    params.require(:event).permit(:title, :description, :category,
+    :course_id, :duration, :start_time, :end_time, :instructor, :location,
+    :id_code, :status, :timecard_ids, :person_ids, :comments, :is_template, :template_id ,
```
mxie: There's an extra space after `template_id`!

app/controllers/events_controller.rb
```diff
@@ -2,97 +2,55 @@ class EventsController < ApplicationController
...
+    :department_ids => [])
```
mxie: Thoughts on using newer hash syntax here?

app/models/event.rb
```diff
@@ -2,17 +2,17 @@ class Event < ActiveRecord::Base
...
 
   validates_presence_of :start_time, :end_time
   validates_chronology :start_time, :end_time
-
+# In order to implement templates, I think we'll want self-referential
+# association. This will require a template_id on the event table.
+# https://collectiveidea.com/blog/archives/2015/07/30/bi-directional-and-self-referential-associations-in-rails
+# http://stackoverflow.com/questions/5109893/rails-how-do-self-referential-has-many-models-work
+  belongs_to :template, :class_name => "Event"
+  has_many :templated_events, :class_name => "Event", :foreign_key => "template_id"
```
mxie: Same thing, re: hash synx

app/models/event.rb
```diff
@@ -116,6 +119,19 @@ def completed?
     status == "Completed"
   end
 
+  def use_template
```
mxie: What do you think about naming this `use_as_template`? Is that what we're doing here?

spec/factories/event.rb
```diff
@@ -1,7 +1,7 @@
 FactoryGirl.define do
   factory :event do
-    title "A Training Event"
-    sequence (:id_code) {|n| "code_#{n}" }
+    sequence (:title) { |n| "Training Event #{n}" }
```
mxie: There's an extra space after `sequence`. Same for the line below.

app/controllers/events_controller.rb
kgf: No, because in line 2 we have a before filter for the authorization. Since the authorization is based on the object the user is trying to access, this before filter loads the object for us.

An example is here
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/controllers/items_controller.rb#L10

app/models/event.rb
kgf: I changed it to `use_a_template`; This is where we're putting a template to use.

mxie: Ohhh I see what you mean. In that case, the original name is probably fine. Or maybe `apply_template.

app/controllers/events_controller.rb
mxie: Got it. Thanks!

mxie (approved):

Issue #396, opened 2017-05-06
In View Person Timecards - there is no column sort


VanAuken:
In View Person Timecards - there is no column sort.  The table shows the entries randomly.

kgf:
There is referring to the person#show action. There are accordion elements on the show page. The bottom one is timecards for that person. This is the table that does not currently have sorting. It should default to sorting by date, with the most recent at the top.

PR #411, opened 2017-07-20
Corrected JS/coffee to add click to sort for timecards

kgf:
Also added the search box back in as a default


kgf:
Addresses #396

VanAuken:
Looked it over as best I am able.

PR #417, opened 2017-07-25
alter columns  on tasks table

kgf:
reorder priority to be after assignees and shortened "Requirements Met" to "Reqs"

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -1,5 +1,7 @@
 class Notification < ActiveRecord::Base
 #  serialize :departments, Array
+  has_paper_trail
+
   has_and_belongs_to_many :departments
   belongs_to :event
   has_many :recipients
--- a/spec/features/notifications_spec.rb
+++ b/spec/features/notifications_spec.rb
@@ -5,8 +5,7 @@ RSpec.describe "Notifications" do
     before (:each) { sign_in_as('Editor') }
     let!(:notification)  { create(:notification, subject: "Howdy from Hopedale") }
     context "basic CRUD" do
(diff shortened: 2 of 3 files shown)


spec/features/notifications_spec.rb
```diff
@@ -3,7 +3,14 @@
 RSpec.describe "Notifications" do
   describe " visit notifications" do
     before (:each) { sign_in_as('Editor') }
-
+    let!(:notification)  { create(:notification, subject: "Howdy from Hopedale") }
+    context "basic CRUD" do
+      let!(:notification)  { create(:notification, subject: "Howdy from Hopedale") }
+      it "should display one at a time" do
+        visit notification_path(notification)
+        expect(page).to have_content("Howdy")
```
masukomi: this only proves that there is at least one Howdy. it doesn't prove that it only displays one at a time.

also the 2nd one should have a different variable name (`let!(:notification_2){...}`) so that no-one mistakenly thinks you're just overwriting the 1st one without creating 2 entries in the db, which you are.

also, this has nothing to do with the "label" (title) so should probably be in a separate pr

app/models/notification.rb
```diff
@@ -1,5 +1,7 @@
 class Notification < ActiveRecord::Base
 #  serialize :departments, Array
+  has_paper_trail
```
masukomi: i don't see `gem 'paper_trail'` in the `Gemfile`  or this PR so i don't see how it will work.

also i don't see what this has to do with the trim label.

masukomi (changes requested):

masukomi (approved):

PR #418, opened 2017-07-25
Require person on assignment

kgf:
Assignment does not make much sense without a person

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -5,6 +5,7 @@ class Assignment < ActiveRecord::Base
   belongs_to :requirement
   delegate :task, :to => :requirement
   delegate :event, :to => :task
+  validates_presence_of :person
 
   STATUS_CHOICES = [ 'New', 'Active', 'Cancelled' ]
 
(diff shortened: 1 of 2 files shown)


app/models/notification.rb
```diff
@@ -1,5 +1,7 @@
 class Notification < ActiveRecord::Base
 #  serialize :departments, Array
+  has_paper_trail
```
masukomi: again, `gem 'paper_trail'` isn't in the gemfile so don't see how this can work. 
also, seems inapplicable to this pr.... but it appears this pr is just built on #417 so...

spec/features/notifications_spec.rb
```diff
@@ -3,7 +3,14 @@
 RSpec.describe "Notifications" do
   describe " visit notifications" do
     before (:each) { sign_in_as('Editor') }
-
+    let!(:notification)  { create(:notification, subject: "Howdy from Hopedale") }
+    context "basic CRUD" do
+      let!(:notification)  { create(:notification, subject: "Howdy from Hopedale") }
+      it "should display one at a time" do
+        visit notification_path(notification)
+        expect(page).to have_content("Howdy")
```
masukomi: see issue in #417

masukomi (changes requested):

app/models/notification.rb
kgf: https://github.com/ReadyResponder/ReadyResponder/blob/development/Gemfile#L32
It seems to be there in the repo...

masukomi: i was looking at master. :/

masukomi (approved):

PR #422, opened 2017-08-06
requirements now skill OR title enforced

masukomi:
requirements should only be allowed to have a skill or a title but not both. The UI enforced this. The code did not.

### DIFF
--- a/app/models/requirement.rb
+++ b/app/models/requirement.rb
@@ -11,6 +11,7 @@ class Requirement < ActiveRecord::Base
   # validates :skill, presence: title.blank?
   # validates :title, presence: skill.blank?
   validate  :valid_title_or_skill
+  # ^^^ guarantees one or the other has been set (not both)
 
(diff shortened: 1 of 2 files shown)


app/models/requirement.rb
```diff
@@ -92,5 +93,9 @@ def valid_title_or_skill
         errors.add(:title, "must have a title or a skill")
         errors.add(:skill, "must have a title or a skill")
       end
+      if !title.blank? && !skill.blank?
```
kgf: Would `if title.present && skill.present?` be better, get rid of the negatives ?

spec/models/requirement_spec.rb
```diff
@@ -28,6 +30,10 @@
       it { should validate_presence_of(:title).with_message(/must have/) }
     end
 
+    # context "if it has a title AND a skill" do
+    #   it
```
kgf: Remove this until you're ready ? There is nothing much here

kgf (changes requested):
Two changes requested

kgf (approved):

PR #423, opened 2017-08-06
added and used Requirement#assignable?(person)

masukomi:
now you can't assign a person to a requirement if they aren't accidentally requirable by that requirement

### DIFF
--- a/app/models/requirement.rb
+++ b/app/models/requirement.rb
@@ -34,9 +34,7 @@ class Requirement < ActiveRecord::Base
     return false if number_filled >= maximum_people
     return false if people.include? person
 
-    return true if !title.nil? && person.titles.include?(title)
-    return true if !skill.nil? && person.skills.include?(skill)
-    false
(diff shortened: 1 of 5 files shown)


app/models/requirement.rb
```diff
@@ -30,6 +30,13 @@ def assignees
     assignments.active.map { |a| a.person }
   end
 
+  def assignable?(person)
+    return false if number_filled >= maximum_people
+    return false if people.include? person
```
kgf, line 35: This is OK for now, but won't work long term. Requirements may span days, and a person may fulfill the same requirement on different days.

kgf:
I have put this on hold, briefly. The code is fine, but once this is in, you won't be able to assign someone without the title. Currently some production datasets do not have all the background info to support this.

app/models/requirement.rb
kgf: @masukomi Would you have time to address this prior to bugging out ?

Issue #427, opened 2017-08-31
Improve choices on notifications


kgf:
When creating a notification, do not show departments that are not listed on the event. Also, none should be checked by default.

When working on this, be aware that we'll probably make the link to an event optional.

kgf:
Closed by #428

PR #428, opened 2017-08-31
427 improve notifications

kgf:
This addresses #427 and improves specs

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -19,6 +19,15 @@ class Notification < ActiveRecord::Base
 
   VALID_STATUSES = STATUS_STATES.keys
   validates :status, inclusion: { in: VALID_STATUSES }
+  validate :notification_has_at_least_one_recipient
+
+   def notification_has_at_least_one_recipient
+     # As we add more ways to choose recipients,
+     # we'll need to expand this validator
+     if departments.blank?
+       errors[:base] << "All recipients can't be blank"
+     end
+   end
 
   def available_statuses
     if status
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -20,6 +20,7 @@ class Notification < ActiveRecord::Base
   VALID_STATUSES = STATUS_STATES.keys
   validates :status, inclusion: { in: VALID_STATUSES }
   validate :notification_has_at_least_one_recipient
+  validates_presence_of :subject
 
   def available_statuses
     if status
@@ -66,7 +67,7 @@ private
     # As we add more ways to choose recipients,
     # we'll need to expand this validator
     if departments.blank?
-      errors[:base] << "All recipients can't be blank"
+      errors[:departments] << "All recipients can't be blank"
     end
   end
 
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -21,14 +21,6 @@ class Notification < ActiveRecord::Base
   validates :status, inclusion: { in: VALID_STATUSES }
   validate :notification_has_at_least_one_recipient
 
-   def notification_has_at_least_one_recipient
-     # As we add more ways to choose recipients,
-     # we'll need to expand this validator
-     if departments.blank?
-       errors[:base] << "All recipients can't be blank"
-     end
-   end
-
   def available_statuses
     if status
       STATUS_STATES[status]
@@ -65,6 +57,16 @@ class Notification < ActiveRecord::Base
       recipients.each do |r|
         r.notify! twilio
       end
+    end
+  end
+
+private
+
+  def notification_has_at_least_one_recipient
+    # As we add more ways to choose recipients,
+    # we'll need to expand this validator
+    if departments.blank?
+      errors[:base] << "All recipients can't be blank"
     end
   end
 
--- a/app/views/notifications/_form.html.erb
+++ b/app/views/notifications/_form.html.erb
@@ -2,10 +2,11 @@
   <h3><%= "#{@event.title} [#{@event.id_code}] "  %></h3>
   <h4>(<%= format_time_span(@event.start_time, @event.end_time, @event.duration) %>)</h4>
   <%= f.error_notification %>
+  <div class="alert alert-danger"><%= f.error :base %></div>
 
   <div class="form-inputs">
     <%= f.association :event,
-                      collection: Event.upcoming,
+                      collection: Event.upcoming + [@event],
                       include_blank: false %>
     <%= f.input :subject, :hint => "Used for text messages" %>
     <%#= f.input :body, :hint => "Not used for text messages" %>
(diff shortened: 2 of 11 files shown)


spec/models/notification_spec.rb
```diff
@@ -5,23 +5,33 @@
...
     end
-
-    it 'has many recipients'
   end
-
   describe 'validations' do
     context 'status' do
       it 'accepts a valid status' do
-        notification = build(:notification, status: 'Scheduled')
+        allow_any_instance_of(Notification).to \
```
corincerami: Instead of this, could you instead make sure in the factory that each notification has at least on recipient?

kgf: I prefer not to. You'll see in other codebases things in the Factory code like ->
    association :owner, factory: :person
But this makes the specs brittle, since you really don't care about the associated model. Also, you're now saving a lot more to the db, slowing tests. 
Imagine the person has two more associations, and one of those has some associations too ! It can, and will, cascade.

kgf: @chrisccerami Thanks for looking at this. If that answers your question, and you don't have other concerns, would you be willing to merge ? I would appreciate it.

Thanks

app/models/notification.rb
```diff
@@ -19,6 +19,15 @@ class Notification < ActiveRecord::Base
 
   VALID_STATUSES = STATUS_STATES.keys
   validates :status, inclusion: { in: VALID_STATUSES }
+  validate :notification_has_at_least_one_recipient
+
+   def notification_has_at_least_one_recipient
```
mxie: The indentation here is a little off, but more importantly - should this be a private method instead?

app/models/notification.rb
```diff
@@ -19,6 +19,15 @@ class Notification < ActiveRecord::Base
...
+     # As we add more ways to choose recipients,
+     # we'll need to expand this validator
+     if departments.blank?
+       errors[:base] << "All recipients can't be blank"
```
mxie: To make it more specific, I think we can do `errors.add(:departments, "<error message>")` instead so that it's clear what's missing.

app/views/notifications/_form.html.erb
```diff
@@ -2,10 +2,11 @@
   <h3><%= "#{@event.title} [#{@event.id_code}] "  %></h3>
   <h4>(<%= format_time_span(@event.start_time, @event.end_time, @event.duration) %>)</h4>
   <%= f.error_notification %>
+  <div class="alert alert-danger"><%= f.error :base %></div>
```
mxie: I _think_ if we do what I mentioned re: `errors.add`, we might not need this. Also, does `error_notification` have the same CSS clases?

app/views/notifications/_form.html.erb
```diff
@@ -2,10 +2,11 @@
...
 
   <div class="form-inputs">
     <%= f.association :event,
-                      collection: Event.upcoming,
+                      collection: Event.upcoming + [@event],
```
mxie, line 8: Could we have `upcoming` include this event?

spec/features/notifications_spec.rb
```diff
@@ -3,26 +3,34 @@
...
       it "should display a single notification" do
         visit notification_path(notification)
         expect(page).to have_content("Howdy")
       end
     end
     context "from an event" do
-      let!(:event)  { create(:event) }
+      let!(:valid_department)  { create(:department, manage_people: true, name: 'Valid Dept') }
+      let!(:invalid_department)  { create(:department, manage_people: false, name: 'Invalid Dept') }
+      let!(:event)  { create(:event, departments: [valid_department]) }
```
mxie, line 19: We used `Event.new` previously. Do we have to create this in the db or can we get away with using `build` instead of `create` for the same effect?

spec/models/notification_spec.rb
mxie: I think instead of this, you could do something like:

```ruby
notification = build(:notification, departments: [build(:department)])
```

That way, nothing has to be persisted.

spec/models/notification_spec.rb
```diff
@@ -5,23 +5,33 @@
...
+                    receive(:notification_has_at_least_one_recipient).and_return(true)
+        notification = build(:notification)
         expect(notification).to be_valid
       end
 
       it 'rejects an invalid status' do
+        allow_any_instance_of(Notification).to \
```
mxie, line 17: Same here as what I said above. But also: I don't think this test actually tests the status. It could be invalid for a different reason...however, I think rewriting these tests can be done separate from this PR.

spec/models/recipient_spec.rb
```diff
@@ -1,7 +1,9 @@
 require 'rails_helper'
 
 RSpec.describe Recipient, type: :model do
-  it "has a valid factory" do
+  it "has a mostly valid factory" do
```
mxie, line 4: Same here re: this test isn't specific enough.

spec/views/notifications/show.html.erb_spec.rb
```diff
@@ -2,16 +2,17 @@
...
       :event => nil,
       :status => "Active",
       :channels => "Email, Text",
+      :departments => [department]
     ))
   end
 
   it "renders attributes in <p>" do
     render
-    expect(rendered).to match(//)
```
mxie, line 14: 👍

app/models/notification.rb
kgf: Perhaps premature optimization, but once I allow users to choose people individually or groups of people, the error won't be on department any more. It will be the fact that you haven't chosen a value from ANY of the fields....
The message itself could be improved though, perhaps 
`Please select someone to receive this notification from departments` and then add `or groups` to that as appropriate...

app/models/notification.rb
kgf: Yes, good call

spec/models/notification_spec.rb
kgf: I did not think that would work. I thought the department would need to be persisted to have an ID to put into the notification. I'll give it a try though !

mxie: Yeah, the departments would need to be persisted in order to have an idea. However, your validation doesn't check for ids - only that `departments` is a non-nil value.

app/views/notifications/_form.html.erb
kgf: No, because we may be sending out a notification about an event in the past. ("Be sure to have all your paperwork regarding Hurricane Harvey turned in by next Friday")

spec/models/notification_spec.rb
kgf: @mxie 's idea worked. PR updated. Thanks !

spec/features/notifications_spec.rb
kgf: I tried it, no joy. Since this is a feature spec, and we're trying to visit the show page, the event must be persisted.

kgf:
@mxie Does this address your concerns adequately ?

PR #433, opened 2017-09-07
added Verify button to timecards

goldenhub77:
Feature Verify button added to timecards.
All functionality and tests pass successfully. Ready for assessment.

### DIFF
--- a/app/controllers/timecards_controller.rb
+++ b/app/controllers/timecards_controller.rb
@@ -1,6 +1,6 @@
 class TimecardsController < ApplicationController
   before_filter :authenticate_user!
-  load_and_authorize_resource
+  load_and_authorize_resource except: :verify
 
   def index
     @timecards = Timecard.all.order(start_time: :desc)
@@ -41,11 +41,11 @@ class TimecardsController < ApplicationController
   end
 
   def verify
+    authorize! :edit, Timecard
     @timecard = Timecard.find(params[:timecard_id])
     if @timecard.status == 'Verified'
       redirect_to request.referrer || timecards_path, notice: 'Timecard already verified.'
     else
-      authorize! :edit, @timecard
       @timecard.status = 'Verified'
       if @timecard.save
         respond_to do |format|
--- a/app/controllers/timecards_controller.rb
+++ b/app/controllers/timecards_controller.rb
@@ -40,6 +40,28 @@ class TimecardsController < ApplicationController
     end
   end
 
+  def verify
+    @timecard = Timecard.find(params[:timecard_id])
+    if @timecard.status == 'Verified'
+      redirect_to request.referrer || timecards_path, notice: 'Timecard already verified.'
+    else
+      authorize! :edit, @timecard
+      @timecard.status = 'Verified'
+      if @timecard.save
+        respond_to do |format|
+          format.html {
+            redirect_to request.referrer || timecards_path
+          }
+          format.json {
+            render json: @timecard
+          }
+        end
+      else
+        render action: "edit"
+      end
+    end
+  end
+
   def destroy
     @timecard.destroy
     redirect_to timecards_path
--- a/app/controllers/timecards_controller.rb
+++ b/app/controllers/timecards_controller.rb
@@ -1,6 +1,6 @@
 class TimecardsController < ApplicationController
   before_filter :authenticate_user!
-  load_and_authorize_resource except: :verify
+  load_and_authorize_resource
 
   def index
     @timecards = Timecard.all.order(start_time: :desc)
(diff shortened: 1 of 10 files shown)


kgf:
@goldenhub77 , thanks for this help! I look forward to taking a look.

app/controllers/timecards_controller.rb
```diff
@@ -1,6 +1,6 @@
 class TimecardsController < ApplicationController
   before_filter :authenticate_user!
-  load_and_authorize_resource
+  load_and_authorize_resource except: :verify
```
kgf: Rather than carving out an exception for verify, what if we added verify to the Ability class here ?
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/models/ability.rb#L35
If you look at line 27 of that file, you can see how custom actions were added to the Person class.
See further comments in def verify

app/controllers/timecards_controller.rb
```diff
@@ -40,6 +40,28 @@ def update
     end
   end
 
+  def verify
+    authorize! :edit, Timecard
+    @timecard = Timecard.find(params[:timecard_id])
```
kgf: If we added Verify to the Ability class as mentioned above, we would not need lines 44-45. Also, I don't think 44 will work the way we want, as it authorizing on the class Timecard, not an instance of timecard. Removing 45 will also eliminate the Code Climate complaint as well.

app/controllers/timecards_controller.rb
```diff
@@ -40,6 +40,28 @@ def update
...
+    if @timecard.status == 'Verified'
```
kgf: I think we should turn this conditional around, and test that a timecard has a status of "Unverified". A timecard in an error state is not eligible to be verified. In fact, only an `Unverified` timecard can be verified. All others need to be hand edited.
The error message could be something like `Timecard not unverified`

app/controllers/timecards_controller.rb
```diff
@@ -40,6 +40,28 @@ def update
...
+        respond_to do |format|
+          format.html {
+            redirect_to request.referrer || timecards_path
+          }
+          format.json {
+            render json: @timecard
+          }
+        end
+      else
+        render action: "edit"
```
kgf: Where `verify`is coming in as an ajax call, is render edit the expected action ?

app/views/timecards/_timecard.html.erb
```diff
@@ -1,8 +1,11 @@
 <tr>
-  <td><%= link_to timecard.status, timecard_path(timecard) %></td>
+  <td><%= link_to timecard.status, timecard_path(timecard), id: "js-timecard-status-id-#{timecard.id}" %></td>
   <td><%= timecard.start_time.strftime('%a %b %d %Y %k:%M') if timecard.start_time %></td>
   <td><%= timecard.end_time.strftime('%a %b %d %Y %k:%M') if timecard.end_time%></td>
   <td><%= timecard.duration %></td>
   <td><%= truncate(timecard.description, length: 24) %></td>
-  <td><%= table_button_link 'Edit', edit_timecard_path(timecard) if can? :edit, timecard %></td>
+  <td>
+    <%= button_to 'Verify', timecard_verify_path(timecard), form_class: 'timecard-verify', id: "js-timecard-verify-btn-id-#{timecard.id}", class: 'js-timecard-verify-btn btn btn-xs btn-success' if (can? :edit, timecard) && timecard.status != 'Verified' %>
```
kgf: Would need to change the conditional here. 
`if (can? :verify, timecard) && timecard.status == 'Unverified'

Also, would it be possible to break this line up ? It is quite long

app/views/timecards/show.html.erb
```diff
@@ -26,6 +26,8 @@
   <b>Duration:</b>  <%= @timecard.duration %>
 </p>
 
+<%= button_to 'Verify', timecard_verify_path(@timecard), form_class: 'timecard-verify', id: "js-timecard-verify-btn-id-#{@timecard.id}", class: 'js-timecard-verify-btn btn btn-success' if (can? :edit, @timecard) && @timecard.status != 'Verified' %>
```
kgf: Same as above, new conditional and multiple lines ?

spec/features/timecards_spec.rb
```diff
@@ -62,4 +62,49 @@
...
+    it "click 'Verify' button for persons timecard from all applicable pages" do
+      @tc1 = create(:timecard, person: @cj, status: 'Unverified')
+
+      visit timecards_path
+      click_button "Verify"
+
+      expect(page).not_to have_button('Verified')
+      expect(page).not_to have_content('Unverified')
+
+      @tc2 = create(:timecard, person: @cj, status: 'Unverified')
```
kgf: I really like the number of tests, but  I think it would be better to break this into three blocks. You could wrap it in a context.

kgf:
@goldenhub77 thanks for your work on this. I did make some comments that I hope you can take a look at.

Thanks for your help !

kgf:
That's awesome, thanks for the quick turn-around ! I'll take a look in the morning.

kgf:
Thanks for your help with this @goldenhub77 !

Issue #293, opened 2016-11-27
Make the Org Chart menu choices dynamic


kgf:
The menu choices should be `All the departments that are active, and that have managing people true`
Obviously, this will require changes in the people_controller as well.

pcai:
I'm happy to take a shot at this if it's still up for grabs.

kgf:
It is, thanks. I'll try to update the ticket with better details.

kgf:
If you look at how the `People` menu is implemented, this may be similar. 
Right now, when you request an orgchart, it only shows one department. It would be better if there were a sub-menu, and you could select which department. If you created a new department, AND selected that it managed people, then that new department would begin appearing as a choice for the orgchart.

The orgchart itself should show each group of `division1` values as a section.

pcai:
Thanks @kgf - I didn't realize the referenced menu didn't exist yet. That sounds straightforward enough though. I opened PRs for a few fixes that were blocking me, will revisit this once those are merged.

kgf:
@pcai Does this still interest you ? Hacktoberfest is coming, I'm sure you need a new t-shirt !

pcai:
Sorry for the (long) delay. I opened a preliminary PR to check if I'm on the right path? Please see attached screenshots in 462.

PR #462, opened 2017-09-28
make orgchart dynamic

pcai:
Resolves https://github.com/ReadyResponder/ReadyResponder/issues/293

Makes the 'reports' menu dynamic, creating an org chart menu item for each active department that manages people:
e

Updates org chart page to take a department id parameter, instead of hardcoding the results for police (or simply all active members of a certain division regardless of their department):
d


app/views/layouts/orgchart.html.erb
```diff
@@ -1,173 +1,128 @@
...
+      .corporal,
+      .senior,
+      .division_heading,
+      .squad_heading {
+        border: 4px solid #555;
+      }
+    </style>
+  </head>
+  <body>
+    <%= yield %>
```
pcai: previously, this layout ignored the actual view template being rendered. I've added a `yield` here to address this.

kgf:
Could we make it a sun menu, have the departments listed to the side?

app/controllers/people_controller.rb
```diff
@@ -14,10 +14,12 @@ def signin
   end
 
   def orgchart
-    @department = Department.where(shortname: "BAUX").first
-    @people = @department.people.active if @department
-    @page_title = "Org Chart"
+    @department = Department.find(params[:dept_id])
```
kgf: This line causes issues for Codacy. I wonder if that's trying to tell us this is really about the department, not about people ?

app/views/layouts/orgchart.html.erb
```diff
@@ -1,173 +1,128 @@
...
-      <li class="squad_heading">Squad 2</li>
-      <% Person.active.division2.squad2.each do |p| %>
-        <%= render :partial => 'coc_boxes', :locals => { :p => p} %>
-      <% end %>
-    </ul>
-  </li>
-</ul>
-<span id="thedate" > <%#= Date::today().to_s %></span>
-</div>
+      .chief,
```
kgf: Not for this PR, but this hardcodes stuff has got to go too.

app/views/people/orgchart.html.erb
```diff
@@ -116,31 +116,30 @@ li.squads {
...
+    <%= render :partial => 'coc_boxes', :locals => { :p => p} %>
   <% end %>
 </ul>
 
-
 <div class="divisions">
 <ul class="division" id="division1" >
   <li class="division_heading">Division 1</li>
-  <% Person.active.division1.squadC.each do |p| %>
+  <% @people.division1.squadC.each do |p| %>
```
kgf: Much better, thanks

app/controllers/people_controller.rb
pcai: It seems to be complaining that I'm not scoping it correctly so that cancancan's ability.rb is enforced: https://github.com/CanCanCommunity/cancancan/wiki/authorizing-controller-actions

i can try `Department.accessible_by(current_ability).find(params[:dept_id])` or something to that effect.

kgf: I was thinking the problem would go away if this were in the Departments controller. It's a report about a department, it should probably be there.

pcai:
So unfortunately, moving the action to the departments controller didnt help with the codacy issue. i will use the cancancan check to address.

kgf:
It may be that you now don't need the @department.find call at all. That was how I was thinking it would resolve that issue.

kgf:
I think that this line will cause @department to be automatically be defined.
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/controllers/departments_controller.rb#L4

pcai:
I got it working with `@department = Department.accessible_by(current_ability).find(params[:id])` but now I know what you mean - that would duplicate the `load_and_authorize_resource` call.

pcai:
...and we got our wires crossed, but on the same page now. One moment while I push that change

pcai:
heres what this looks like now with a menu on the sidebar (was this what you mean @kgf ?):

The topnav entry for `Org Chart` can be reverted to a single menu item that defaults to the first eligible department, or it can go to `/orgchart` with no `id` defined which I suppose can lead to an index page of sorts. Not sure what is preferred here.

a1
a2

spec/controllers/departments_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    # catch view rendering errors
+    render_views
+    
+    it "shows an orgchart" do
+      get :orgchart, {id: department.id}
+      expect(response).to be_success
+    end
+    
+    context "as a reader" do
+      before(:each) { login_as('Reader') }
```
pcai, line 19: I wasn't sure which roles should have access to the department orgcharts so I arbitrarily picked `Reader` as a role without permission, for illustrative purposes.

app/assets/stylesheets/orgchart.scss
```diff
@@ -0,0 +1,141 @@
...
+  
+  li.active {
+    font-weight: bold;
+    background: #ccc;
+  }
+}
+
+@media print {
+  .orgchart_menu {
+    display:none;
```
pcai, line 22: feels a bit weird to put this media query inside a css file explicitly marked for `screen`, but I gather the orgchart page was going through hoops to remain print-friendly and thought I should carry that tradition forward. Everything is compiled into application.css with a `require_tree .` so I didnt have much flexibility to spin up a new print stylesheet without re-arranging a bunch of stuff.

kgf:
I was intending to leave the page itself fairly clean. You are correct, it is usually printed out. A better verison would have media type to control what is printed.
What I was thinking of was the layout of the menus themselves. I see you're using firefox. An example would be when, in firefox, you go to `View - Toolbars`, and the next set of choices come out to the side. It would be `Reports - Orgcharts` and then the active departments would be off to the side.
Make sense ?

pcai:
Oh. OK! Let me see if I can play with bootstrap to get it to do that.

pcai:
Hows this? Haven't tested but probably not mobile friendly. I can either keep the navigation links in the orgchart layout or revert it.

a1

kgf:
I like this, with the fly out menu. While perhaps not mobile friendly, neither is the report.

PR #468, opened 2017-09-29
Ui v3

acenasir:


kgf:
This is not a good thing, sorry. You're trying to merge a feature branch (that is not complete) into development. We're not ready for that yet, and those making the code changes will make that PR. Perhaps you could take a look at the issues, and choose something from there ?

Issue #409, opened 2017-06-30
When looking at Titles, the show form does not show the skills


kgf:
You need to go into the edit screen to see which ones are selected.

PR #474, opened 2017-10-01
Shows skills at Title show page

mertbulan:
#409


kgf:
This certainly works, and I will accept it if needed. But some of the titles can be quite long, and will not necessarily break at the comma.
Would you be willing to convert this to an unordered list?
Thanks for your help.

PR #475, opened 2017-10-01
refactored events.js.coffee, dataTables(), split dropdown for events on top menu

goldenhub77:


kgf:
Wow, you've done a lot, thanks.

There seems to be an issue with the filtering. I have a completed event from June of this year ( And yes, it's nice having the year !!) It does not show when I check recent. It does show in Archives.

I am wondering if we can close the gap between the word events and the drop arrow. It seems much wider than the samples. If not, we should probably change the background from black so that they show as distinct entries.
screen shot 2017-10-01 at 10 12 14 pm
screen shot 2017-10-01 at 10 15 44 pm

config/routes.rb
```diff
@@ -83,12 +83,20 @@
...
+  get 'events/archives', to: 'events#archives'
+
   resources :events do
     resources :tasks, only: [:new, :create]
     resources :availabilities
     resources :notifications
   end
 
+
+
```
kgf: Thsi seems ..... odd

app/assets/javascripts/events.js.coffee
```diff
@@ -1,36 +1,41 @@
 jQuery ->
   $("#event_title").focus()
-  $table = $('#events').dataTable
+  $eventTable = $('#events').dataTable
+    'dom': "fr<'event-table-info't>ip",
```
kgf, line 4: fr<'event-table-info't>ip
What is the `t` just inside the `>ip` ?

goldenhub77: it stands for table. All of those letters stand for a particular section of dataTables() and how it renders the table itself. It means it will add a div called 'event-table-info' inside of the table when compiling the html.

goldenhub77: <div class='event-table-info'>
</div>

kgf: I expected the `t` to be outside the `>`, but I could easily be wrong.
I was looking at this page
https://datatables.net/reference/option/dom

goldenhub77: If you put the t on the outside it will not render the div in the proper location. I can get away with having to do this by using a really long selector to append to but I was trying to avoid that.

kgf: Sorry, I should have been clearer. I was not suggesting another approach, just trying to learn !

kgf:
I'm going to merge this, you can push to the same branch for another PR as needed.

PR #458, opened 2017-09-25
Changed ID to title field in the events#show/availabilities view

ghost:
#454

PR #481, opened 2017-10-02
#458 fix

ghost:
#458 Replaced unhelpful "id" field in availabilities listing with the person's title. Adjusted datatables to automatically hide and sort on the title order.


app/assets/javascripts/availabilities.js.coffee
```diff
@@ -0,0 +1,6 @@
+jQuery ->
+  $('#availabilities').dataTable
+    order: [[0, 'desc'], [1, 'asc']],
+    columnDefs: [
```
kgf, line 4: Is this enough for `columnDefs` ? While we have a hidden column now, I believe clicking on the column marked `Title` will still sort by the title column. I think you need something like => `{ orderData: 2, targets: 3 }`
You can see that here ->
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/assets/javascripts/events.js.coffee#L5

ghost: I think you are right. orderData , targets will allow sorting by rank when user clicks title column (instead of the title word(s) itself), worthwhile if there is no use case for sorting by title names

Issue #485, opened 2017-10-02
On person#edit, should be able to add additional phones


kgf:
Currently, you can not add additional ways to contact people. There is room for one phone, and one email. In fact, the backend structure supports unlimited channels. 
The solution is to give people the ability to add a new channel beyond the existing two.

Fell free to ask questions.

SGospodinov:
I'll be happy to take work on this one. I have the following questions:
1. Can I use additional gems namely https://github.com/nathanvda/cocoon?
2. Do you need this feature to be tested?

kgf:
Using that Gem is OK.
Tests are always good !

Thanks for your help !
Kevin

Issue #483, opened 2017-10-02
Placeholder text gives wrong phone number format


ghost:
Form validates numbers with a + (country code) prefix, e.g. in US +15555555, but placeholder is not in this format.

tegon:
Hi! Can I tackle this one?

kgf:
@tegon Sorry, there is already a PR in for it. If there is something specific you'd like to work on ( Front-end back-end ) I can probably point you to an issue.
Thanks for your interest
Kevin

tegon:
@kgf Sure thing, what do you have open for backend?

Issue #482, opened 2017-10-02
New Person form, in phone information, label should be 'number' not 'content'


ghost:
In phone info, label for number says “content” . Should be number.

kgf:
Any solution for this needs to NOT change the label for email.

kgf:
There is code here 
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/views/people/_channel_fields.html.erb
But I'm not sure it's still in use. There is also
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/views/channels/_form.html.erb
that I think is in use, by both the phone and the email.

SGospodinov:
@kgf, I believe 5b55eab resolves this one as well.

kgf:
@SGospodinov Thanks for your help. Yes, this issue is resolved

Issue #450, opened 2017-09-21
Upgrade from Rails 4.2 to Rails 5.0


kgf:
We are currently on Rails 4.2. There are features we would like to develop that would be a lot easier using Action Cable and other features available in Rails 5. Therefore, we would like to upgrade.

Since I don't expect anyone to do the entire ticket, I have created a branch for this effort.
`upgrade_from_rails_4_to_rails_5` Any progress should be PRed against this branch.

A good place to start.
http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-2-to-rails-5-0

Thanks for your help.

kgf:
This may also prove useful.
http://www.ready4rails.net/
https://stackoverflow.com/questions/38442248/how-to-upgrade-from-rails-4-2-6-to-rails-5-0

phsacramento:
Doing

kgf:
Thanks for your help @henriquesacramento

sa77:
@kgf i can help out on this if it is not being worked on.
@henriquesacramento are you currently working on this ?

kgf:
@sa77 We have not heard back from Paulo yet. Could you please start on this ? It is actually probably the highest priority issue, as there are some other tickets I will create once this is complete.
Please update this issue, so that if Paulo is also working on it, we can coordinate.
Thanks for your help.
Kevin

phsacramento:
@kgf I'm doing this. 

Update:
the update has broken some dependencies, I'm working on it.

kgf:
Ah, that's great @henriquesacramento 
Thanks for the update.
As mentioned above, there is a feature branch for any partial PRs -> `upgrade_from_rails_4_to_rails_5`

phsacramento:
@kgf Now we need use strong parameters in project, the gem protected_attributes in no longer compatible with rails 5.
So change all controllers in entire application is need

Update:

@kgf and fix all tests suite

phsacramento:
@kgf Now in rails 5 belongs_to association is required, is not longer possible create record without they respective associations. I will add a fallback for that feature.

ex: 

```
belongs_to :contact, class_name: "Person", optional: true
```

Make sense?

phsacramento:
![captura de tela 2017-10-03 as 12 55 04](https://user-images.githubusercontent.com/2622863/31135029-3905751c-a83a-11e7-81ad-e9dd115a6a83.png)

kgf:
[...]

PR #486, opened 2017-10-02
Adding multiple emails and phones for person #485

SGospodinov:


kgf:
Thanks for your help, @SGospodinov !
Was upgrading the version of Rails a requirement ? I would prefer that to be in a separate PR, to make it more apparent.

SGospodinov:
@kgf, it wasn't. I did it by mistake. Do you want me to change it and rebase the commit?

app/views/phones/_phone_fields.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="nested-fields">
+  <fieldset>
+    <%= phone.input :id, :as => :hidden %>
+    <%= phone.input :category, :as => :select, :collection => Channel::CATEGORIES %>
+    <%= phone.input :content, :placeholder => '9785551212' %>
```
kgf: Could you please change this to "number" ? See issue #482 
Also, the placeholder text should be +19995551212 ? See issue #483 
Thanks !

kgf: If you could revert the Rails upgrade, and the two other small changes, that would be great. Thanks for your help!

kgf:
Thanks for those changes, the phone looks good now. But we still have the Rails upgrade.

kgf:
Whoa, that broke a few tests !!

SGospodinov:
@kgf, now everything works fine. The gems version change was causing the problem. The rails update is opened as a separate issue (#450), so we could discuss it there.

Issue #480, opened 2017-10-02
Inspection form should allow you to choose an inspector


kgf:
This is a follow-up to #94

kgf:
@ChanChar : Need another one ?

ChanChar:
Sweet. Will tackle this one as well.

PR #478, opened 2017-10-02
Change `inspections` association to `inspectors`

ChanChar:
Resolves https://github.com/ReadyResponder/ReadyResponder/issues/94

Changes:

  * Adds `person` association to `Inspection`
  * Adds `inspectors` association to `Person`

PR #490, opened 2017-10-03
Add Inspector Option to Inspection Forms

ChanChar:
Resolves https://github.com/ReadyResponder/ReadyResponder/issues/480

Changes:

  * Adds `inspector` field to any inspection forms
  * Fix seeds entry for `Grant`
  * Add inspection controller specs
  * Update `inspection` factory to build associations
  * Adds `login_manager` spec helper method

Examples

screenshot 2017-10-02 19 36 01
screenshot 2017-10-02 21 22 30


spec/factories/inspection.rb
```diff
@@ -1,5 +1,7 @@
 FactoryGirl.define do
   factory :inspection do
+    person
+    item
```
kgf, line 4: I would prefer not to use this style. I have found it results in cascading lines of objects bring created. Imagine if the person factory then has a reference to an event, that then has take and assignments and so on. I prefer to either create them explicitly in the spec, or stub them.
Thanks for your help

kgf:
How dies this relate to your work on #478 ? I would like to use inspectors whenever possible.
Other than these two small issues, looks good. Thanks !

PR #496, opened 2017-10-03
Merge Dev to Ui_v3

goldenhub77:
Has known rspec failures, PR for assessment.


kgf:
I would also be interested in seeing goldenhub77:ui_v3 into ReadyResponder:development, which would trigger a run of the Travis test suite

kgf:
We're going to try something else...

PR #499, opened 2017-10-04
Added last editor to events and tasks.

### LINKED ISSUE #498: Add last editor to event show
We'll want this on all things eventually, but event show for now.

kgf:
This resolves #498

### DIFF
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -1,7 +1,7 @@
 class TasksController < ApplicationController
   before_action :authenticate_user!
   load_and_authorize_resource
-  
+
   before_action :set_event, only: [:new, :create]
   before_action :set_task, only: [:show, :edit, :update, :destroy]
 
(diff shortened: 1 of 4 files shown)


app/controllers/tasks_controller.rb
```diff
@@ -10,6 +10,8 @@ def index
   end
 
   def show
+    @page_title = @task.title
+    @last_editor = last_editor(@task)
```
ChasManRors, line 14: Out of curiosity, what is suppose to happen when `Task.find(params[:id])` can not find a task?

kgf: There would be nothing to show. @task itself is set by the authorization framework. If there is no task, the view will have all kinds of issues.
Here is the def of `last_editor`
https://github.com/ReadyResponder/ReadyResponder/blob/b38950c24f73ae3075107b3cf9d4c26122f2d266/app/controllers/application_controller.rb#L19

ChasManRors: I put my comment one line too low.

I thought @task was set by the `before_action :set_task`. Is it suppose to be set by the authorization framework? That does not appear to be what is happening
  
```
    def set_task
      @task = Task.find(params[:id])
    end
```

PR #505, opened 2017-10-04
Changes uploaded files' filename to a random token

### LINKED ISSUE #429: Improvements to file upload security
The following is a quote from @houhoulis in issue #306. I would like to implement this, but #306 will be closed soon, and I did not want to loose track of this.

> I'm not sure where you saw this behavior, or how many filenames you tested with. But I can see a routing error locally if I use a file with dots in its name, such as "Pic 2017-01-01 at 10.10.10am.png". With the routing error, I see the filename in the browser rather than the picture, as you reported.
> 
> This appears to be Rails behaving as designed. See
> http://guides.rubyonrails.org/routing.html#dynamic-segments
> 
> One fix I found on Stackoverflow is to change the route declaration to allow dots in the image filename:
> 
>   get '/uploads/person/portrait/:id/:basename.:extension', controller: 'people',
>     action: 'download', type: 'portrait',
>     constraints: { basename: /[^\/]+/ }  # <--- this "constraints" part would be new
> But I'd rather handle filenames a bit more carefully, for a couple reasons. One is it feels sort of creepy to show the person's original filename in everyone's browser network activity. Another is I worry we might hit additional edge cases where the filename causes problems in Rails. Including possible hacking attempts, where someone attempts to get ReadyResponder to misinterpret malformed filenames. (The Carrierwave README mentions this briefly.)
> 
> I think Carrierwave might have an option to generate a filename based on the hash of the contents. If we can make use of a Carrierwave feature like that, that feels smarter than dancing around edge cases ourselves. If not, then (from the README) at least we could set the sanitize_regexp = something strict like /[^[:word:]\-]/ (strip out everything from filenames except the letters, numbers, underscores, and dashes)

cesc1989:
This PR fixes #429 by setting a random token as the filename.

When a file would have many dots in its name (`Pic 2017-01-01 at 10.10.10am`), the file URL would broke rendering a broke image thumbnail. One workaround could be to use a regexp to sanitize filenames but it just doesn't feel right (to me and this is also suggested in [Carrierwave's wiki](https://github.com/carrierwaveuploader/carrierwave/wiki/How-to%3A-Create-random-and-unique-filenames-for-all-versioned-files) for different purposes, though.)

I also took the job to remove unnecessary comments in `*_uploader.rb` files so the `file_name` method is visible at first glance.


kgf:
@cesc1989 Just to let you know. I'm going to want to pull this and test it more throughly, given the possible impact. I won't get to that until tonight.

cesc1989:
No problem.

kgf:
Thanks for your help, @cesc1989 ! This is very helpful.
Fixes #429
Kevin

cesc1989:
You're welcome. My pleasure to contribute to this project.

Issue #503, opened 2017-10-04
Add created_at, edited_at and last editor to notification#show


kgf:
You can look at item_category#show for an example of what is being asked for.

kgf:
@tuwukee is this one of interest ?

kgf:
@bother7 is helping with this one.

PR #517, opened 2017-10-05
issue #503.

bother7:
added created_at, edited_at, and last editor to notification#show. the <p> is below the rest of the notification information, but above the recipient table


kgf:
@bother7 thanks for your help. 
Could you please make two changes.
1 yes, the block should be below the other elements. If you look at item category show view, that's what we're looking for. I would like it to be consistent.
2 could you please not include gemfile.lock. You're using an experimental version of bundler that others are unlikely to have yet.

kgf:
Looking at this PR may be helpful.
https://github.com/ReadyResponder/ReadyResponder/pull/499/files

kgf:
Sorry,@bother7,  my earlier comment was unclear. I did not want you to delete Gemfile.lock ( as the Rails app can't run without it), but to simply not include it in your pull request.
And the view file is better, but I would like to drop it even further, below the recipients table.
Thanks for your help,
Kevin

bother7:
Hey, thanks for your patience! First time doing one of these. I think there are some validations that are acting wonky on events and notifications, so I was having some trouble getting test data running.

kgf:
Oh no problem, very friendly project...
Sadly, removing `Gemfile.lock` is still part of the PR. 
Also, we don't want to add Gemfile.lock to the `gitignore` file. Typically it *would* be checked in. In this case, I'm choosing not to include your version, because the version of bundler is pre-release beta software.

bother7:
sure np! Just a heads up, if it's because of rmagick, there are some compatibility issues between the rmagick gem and macOS Sierra.

kgf:
This has been replaced by a later PR. #521

Issue #506, opened 2017-10-04
NoMethodError on text notification send


ghost:
![notification_nil](https://user-images.githubusercontent.com/21265184/31189451-49ab7070-a906-11e7-965c-ba50249570d8.jpeg)

Rails error on submitting a new notification to be sent out via text. @client seems to be nil. Maybe my local environment is missing Twilio credentials? Are there configuration instructions that need to be added to "Getting Started" to make this work locally?

kgf:
We should add a guard to check for a `nil` client. 
Will need to research what the possible solutions might be...

PR #518, opened 2017-10-05
When a valid client can not be initialized, throw an error on delivery

yez:
An error is thrown and caught via the NotificationsController to be more
friendly than a "NoMethodError"

Specs were added to the notifications#create method as well.

Addresses #506

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -34,7 +34,12 @@ class NotificationsController < ApplicationController
     @notification = Notification.new(notification_params)
     if @notification.save
       if @notification.status == "Active"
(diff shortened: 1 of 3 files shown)


kgf:
Nice work, @yez Thanks.
Can I interest you in another ?

yez:
@kgf I'll see what I can do, what did you have in mind?

kgf:
@yez, That would be good, thanks.
What issue would depend on where your interests lie. I have ones for front JS stuff, and others that are more backend, and others that are straight Rails. There are even issues/enhancements that have not been written up yet...
I really appreciate people's help, so I want to try to give issues that interest them.

PR #520, opened 2017-10-06
update info <id_code> text response template to include assignments

### LINKED ISSUE #514: Text to info should show next assignment
When an end user texts `info idcode` currently, they get generic details on the event. It would be good if they got these details, and also their personal assignment, assuming they have one. It should be the next assignment that is current, meaning the **end_time** for the assignment is > now.

Any questions, please ask.

sa77:
resolves #514 

image

### DIFF
--- a/app/views/msg/info.erb
+++ b/app/views/msg/info.erb
@@ -1,13 +1,19 @@
-Event Name: <%=event.title%> (<%= event.id_code %>)
-Duration: <%= event.start_time %> to <%= event.end_time %>
+Name: <%=event.title%> (<%= event.id_code %>)
+<%= event.start_time %> to <%= event.end_time %>
 Description: <%= event.description %>
 <%
   avp  = event.available_people
   pavp = event.partially_available_people
   uavp = event.unavailable_people
   lame = event.unresponsive_people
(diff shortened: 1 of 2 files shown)


app/views/msg/info.erb
```diff
@@ -1,13 +1,19 @@
...
   uavp = event.unavailable_people
   lame = event.unresponsive_people
+  short_date_time = lambda { |from, to| "#{from.strftime('%m/%d %H%M')}-#{to.strftime('%H%M')}" }
 %>
 Available: <%= avp.count %>
 Partially Available: <%= pavp.count %>
 Unavailable: <%= uavp.count %>
 Unknown: <%= lame.count %>
+<% if event.people.include?(person) %>
+Your Assignments: <% event.assignments.each do |asmt| %>
```
kgf: What if we changed this to
`event.assignments.where(person: person).each`
and then we wouldn't need the `if` later. That seems better than iterating over assignments we don't want.

sa77: it took a while to get around model associations .. little confusing with the delegations

and yes that sounds good .. i'll push a commit with the update

sa77:
@kgf PR updated

PR #521, opened 2017-10-06
one last try at issue #503

bother7:


kgf:
As Sia would say "Don't give up !" 
This looks much better. No Gemfile, The requested change is in the right place. We'll see what Travis says...

bother7:
Awesome. Thanks!

Issue #515, opened 2017-10-05
On event#show availabilities, it should be short title, not full title


kgf:
This provides more room for more relevant info, and fits smaller displays better. 
For bonus points, the `Unknown` table below should have the same change from ID to short title.

kgf:
See issue #454

dusty-wil:
Hi @kgf, I'd be interested in looking at this, if that's OK. I'm pretty new to ruby, but am interested in learning.

kgf:
That would be great, @dusty-wil . We're happy to help. 
Post here if you have questions.

dusty-wil:
@kgf Ok, thanks for letting me contribute on this! I think I've got it, but I just want to make sure I'm on the right track. 

This is referring to the event availabilities tab, and the tables of people who are available to respond to an event, and the table of people who haven't yet indicated that they're available (No Response). If that's the case, I think I've got the changes ready.

kgf:
That is all correct

dusty-wil:
@kgf Ok, I was about to make the PR, but I noticed when I run rspec, I get some failures. These seem to be due to the version of chrome I'm using (my dev environment is on a ubuntu server):

Should I try to address these before I make the PR?


Failures:
  1) Events search finds no results
     Failure/Error: visit new_user_session_path

     Selenium::WebDriver::Error::UnknownError:
       unknown error: unrecognized Chrome version: HeadlessChrome/61.0.3163.100
         (Driver info: chromedriver=2.7.236843,platform=Linux 4.4.0-87-generic x86_64)
      /var/lib/gems/2.3.0/gems/selenium-webdriver-3.5.2/lib/selenium/webdriver/remote/response.rb:71:in `assert_ok'
      /var/lib/gems/2.3.0/gems/selenium-webdriver-3.5.2/lib/selenium/webdriver/remote/response.rb:34:in `initialize'
      /var/lib/gems/2.3.0/gems/selenium-webdriver-3.5.2/lib/selenium/webdriver/remote/http/common.rb:83:in `new'
      /var/lib/gems/2.3.0/gems/selenium-webdriver-3.5.2/lib/selenium/webdriver/remote/http/common.rb:83:in `create_response'
      /var/lib/gems/2.3.0/gems/selenium-webdriver-3.5.2/lib/selenium/webdriver/remote/http/default.rb:107:in `request'
[...]

PR #522, opened 2017-10-06
#515 - Show short title instead of full title in tables on event availability page

dusty-wil:
In reference to issue #515, changed instances of ID and Title in Availability and No Response tables to show person.shortrank instead.

### DIFF
--- a/app/views/availabilities/_availabilities_table.html.erb
+++ b/app/views/availabilities/_availabilities_table.html.erb
@@ -16,7 +16,7 @@
       <!-- Full available -->
     <% (availabilities + partial_availabilities).each do |a| %>
       <tr <%== availability_status_class(a, @event) %>>
-        <td><%= Person::TITLE_ORDER[a.person.shortrank] %></td>
+        <td><%= Person::TITLE_ORDER[a.person.title] %></td>
         <td><%= link_to availability_status_label(a, @event), a %></td>
         <td><%= a.person.shortrank %></td>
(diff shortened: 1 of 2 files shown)


app/views/availabilities/_availabilities_table.html.erb
```diff
@@ -16,9 +16,9 @@
       <!-- Full available -->
     <% (availabilities + partial_availabilities).each do |a| %>
       <tr <%== availability_status_class(a, @event) %>>
-        <td><%= Person::TITLE_ORDER[a.person.title] %></td>
+        <td><%= Person::TITLE_ORDER[a.person.shortrank] %></td>
```
kgf: I think this will introduce a bug in click to sort. This column is hidden, and is used for a target for sorting. This line is looking into that hash using the key. The keys are title, not short rank, so it will never find the corresponding value.

kgf:
Thanks for your help. Small change to revert.

dusty-wil:
@kgf got it... do I need to make another PR to fix the issue?

app/views/availabilities/_availabilities_table.html.erb
dusty-wil: @kgf got it... sorry, a little new to Git here. Do I need to make another PR with the changes to revert that issue, or can I do that through Github?

kgf:
Nope, since you pushed to the same branch on your fork, github updated this PR. 
Thanks for your help !

dusty-wil:
great! thanks for your help!

PR #527, opened 2017-10-07
Timecards Table Sort 

radvc:
Reference for this issue https://github.com/ReadyResponder/ReadyResponder/issues/396. I just added data table class in timecards table and create scope in timecard model to order most recent records by start_time.


kgf:
@radvc thanks for your help.

radvc:
@kgf always welcome.

PR #529, opened 2017-10-07
Edit availability includes original person

### LINKED ISSUE #360: Edit Availability does not include original person when being edited
Edit Availability returns the time/dates but not the person if the person's status had changed.
This will affect cases of editing history after a person's status changes.

nesteves:
Fixes #360 

The @availability.person is now always included in the person collection used to populate the select input on the availabilities/_form partial.
Add specs for the availabilites controller #new and #edit actions.
Include child factories that set up active and inactive people.


kgf:
@nesteves Thanks for all your help !

Issue #528, opened 2017-10-07
Validates person has a department


VanAuken:
Everyone needs to belong to a department. 
Ideally, if there is only one department, creating a new person should default to that department

amirkarimi:
Can I work on this issue?

VanAuken:
That would be great ! Thanks @amirkarimi

amirkarimi:
Done https://github.com/ReadyResponder/ReadyResponder/pull/530

ivanfrias:
@VanAuken  please close this so that we all know which ones are still available.

kgf:
@ivanfrias I expect we will close this today. 
However, we do use the `Help Wanted` and `Claimed` tags on the issues to help with knowing what's available.
Thanks for your interest in our project !
Kevin

PR #530, opened 2017-10-07
Add validation to person department to be present

amirkarimi:
And also set the first department as default when there is only one department available.

#528


app/views/people/_form.html.erb
```diff
@@ -14,9 +14,10 @@
     <%= f.input :end_date, as: :string,
                 input_html: { class: 'datepicker', size: 11,
                               value: format_date_value(f.object.end_date) } %>
-    <%= f.label :department_id %>
-    <br>
-    <%= f.collection_select :department_id, Department.managing_people.all, :id, :name, include_blank: true %>
+
+    <%= f.input :department_id, :collection => Department.managing_people.all,
+                :required => true, :selected => Department.managing_people.all.length == 1 ? Department.managing_people.all[0].id : nil %>
```
kgf: This works, but would it be better to get some of this logic out of the view ? Also, this will fire even on edit. I don't think that's a good idea. I would rather only have the default value on new. We may enter people from a department, then turn it back to not managing_people. It's an edge case, but one I expect....

kgf:
@amirkarimi Thanks for your help !
See my comments regarding the default value.

app/views/people/_form.html.erb
amirkarimi: Ah, yes, that's true. I'll work on it.

amirkarimi:
@kgf Addressed the department default value setting.

app/models/person.rb
```diff
@@ -61,6 +61,15 @@ class Person < ActiveRecord::Base
   STATUS = ['Leave of Absence', 'Inactive', 'Active', 'Applicant','Prospect','Declined']
   DEPARTMENT = ['Police', 'CERT', 'Other']
 
+  after_initialize do
```
kgf: Rather than changing the model, what if we changed the controller instead ? 
Adding 
```
    if Department.active.managing_people.count == 1
      @person.department = Department.active.managing_people.first 
    end
```
to here =>
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/controllers/people_controller.rb#L87

app/models/person.rb
```diff
@@ -61,6 +61,15 @@ class Person < ActiveRecord::Base
...
+    if self.new_record?
+      # values will be available for new record forms.
+      if self.department.nil? && Department.managing_people.count == 1
+        self.department = Department.managing_people.first
```
kgf: This will need `Department.active.managing_people.first`, otherwise we may get an inactive department.

app/models/person.rb
amirkarimi: Totally agree with moving this to the controller but why not choosing the department for the model on `new` action when there is only one active department so that the user can see the selected department in the form. If he/she receive the validation error if select the empty department.

amirkarimi: @kgf I've moved this code into the controller `new` action (during page loading). Please let me know if you want it to be on `create` action (during insertion).

app/models/person.rb
amirkarimi: Done

goldenhub77:
@amirkarimi FYI, there are multiple instances of variables being assigned with no purpose within person_new_spec.rb

app/controllers/people_controller.rb
```diff
@@ -55,6 +55,12 @@ def new
     @person = Person.new(status: cookies[:status], state: 'MA')
     @person.emails.build(category: 'E-Mail', status: 'OK', usage: '1-All')
     @person.phones.build(category: "Mobile Phone", status: "OK", usage: "1-All")
+
+    departments = Department.active.managing_people
+    if @person.department.nil? && departments.count == 1
```
kgf: Do we need the `@person.department.nil?` test ? Could it have a value at this point, as we just created it.

kgf:
@amirkarimi, I now you've been at this a while, thanks for your perseverance. I will merge it as is, but wondered if you'd be willing to resolve the Codacy issues ?

amirkarimi:
@kgf Sure, I would like to make this PR perfect. It's been a long time since last time I've written Ruby code.

kgf:
@amirkarimi thanks so much for your help and persistance.

PR #531, opened 2017-10-08
filter upcoming text response to return only relevent events

### LINKED ISSUE #375: text command upcoming should only show events appropriate for viewer
Currently the text message 'upcoming' to the system returns all future events regardless of whether the person is not eligible for working the event.
Example a CERT person will see events which should only be shown to MRC personnel.

sa77:
resolves #375

### DIFF
--- a/app/services/msg/upcoming.rb
+++ b/app/services/msg/upcoming.rb
@@ -3,28 +3,26 @@ class Msg::Upcoming < Msg::Base
   def respond
     events = Event.upcoming.limit(10)
     response = ""
-    events.each do |event|
-      if event.departments.include?(@person.department)
-        if event.people.include?(@person)
-          response += "🔷 "
-        elsif event.available_people.include?(@person)
-          response += "✅ "
-        elsif event.partially_available_people.include?(@person)
-          response += "⚠️ "
-        elsif event.unavailable_people.include?(@person)
-          response += "❌ "
-        else
-          response += "? "
-        end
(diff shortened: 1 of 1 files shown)


app/services/msg/upcoming.rb
```diff
@@ -4,25 +4,27 @@ def respond
     events = Event.upcoming.limit(10)
     response = ""
     events.each do |event|
```
kgf: What if instead we changed this line to `@person.department.events` ? Then we don't need the conditional later ?

kgf:
@sa77 Thanks for your help. Info requested.

app/services/msg/upcoming.rb
sa77: nice .. that works better

sa77:
@kgf PR updated

kgf:
@sa77 Thanks for all your fine work ! The texting section is much better now.
FYI, the emojis are not working _quite_ as well as I'd hoped ( although people really like having the info ), because the colors don't show properly. On a text message, the background color of the bubble varies from person to person. So the color of the emojii changes too !
Anyway, thanks a lot again.
I hope to see more PRs from you in the future.
Let me know if there is another area thta interests you
Regards,
Kevin

sa77:
@kgf regarding SMS emojis, i think the handsets behave differently (maybe with the bubble background, user's personal settings, mobile OS, etc). 
Your project management, quick response to comments, PR review skills are amazing. Keep up the good work. I look forward to contributing more to the codebase. Cheers !

kgf:
Thanks, I appreciate your comments !

PR #536, opened 2017-10-08
people split dropdown and datatable refactor

goldenhub77:
Converted people dropdown to split dropdown
reverted the way tables are rendered on the views for readability (had to update events, people and timecards views)
added authorization check to add button on applicable views
altered and added tests for new functionality


kgf:
How difficult would it be to get the two groups of checkboxes together ? While having the word People as a caption is good semantic markup, it's not a requirement.
Actually, looking at this, we should probably add the word people !
screen shot 2017-10-08 at 6 27 10 pm

app/assets/javascripts/applicants.js.coffee.erb
```diff
@@ -0,0 +1,45 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://jashkenas.github.com/coffee-script/
+jQuery ->
+  $applicantTable = $('#applicants').dataTable
+                      columnDefs: [
+                                    { orderable: false}
```
kgf, line 7: Does columnDefs need a target ? I know the `columns` style doesn't, it based on position.
https://datatables.net/reference/option/columnDefs

app/assets/javascripts/datatables_defaults.js
```diff
@@ -3,6 +3,7 @@
 /* https://datatables.net/reference/option/ */
 
 $.extend( true, $.fn.dataTable.defaults, {
+  "dom": "fr<'event-table-info't>ip",
```
kgf: This file is for all datatables, isn't it ? Seems odd to have something specific to events here.

app/controllers/people_controller.rb
```diff
@@ -147,6 +115,15 @@ def destroy
   end
 
   private
+
+  def manage_people_depts
+    @available_departments = Department.where(manage_people: true)
```
kgf: I think we'd need `Department.active.where....`

spec/features/person_creation_spec.rb
```diff
@@ -13,36 +13,31 @@
...
-    click_link "People"
-    click_link "New Person"
+  scenario "from the people page a.k.a all active", js: true do
+    visit people_path
+    find('.add-btn').click
     fill_in "First Name", with: "John"
     fill_in "Last Name", with: "Doe"
     fill_in "Date of birth", with: "1990-05-13"
     select "Active", from: "person_status" #"Status" also matches Channel status
+    page.execute_script "window.scrollBy(0,1500)"
```
kgf, line 23: Didn't know about execute_script, thanks...
@ThomasCarlson I wonder if this would have helped you ?

spec/features/person_displays_spec.rb
```diff
@@ -18,126 +19,100 @@
...
+    it "returns page for All Active people", js: true do
       visit people_path
-      expect(page).to have_content('People')
+      # Active checked by default
+      expect(page).to have_content('People') # This is in the nav bar
+      expect(page).to have_content('All Active')
+      expect(page).to have_content('Active:')
+      expect(page).to have_content('On-leave:')
+      expect(page).to have_checked_field('Active')
+      expect(page).to have_unchecked_field('On-leave')
```
kgf, line 35: Nice

spec/features/person_displays_spec.rb
```diff
@@ -18,126 +19,100 @@
...
-      expect(page).to have_content('CERT')
-      expect(page).to have_content('Home') # This is in the nav bar
-      expect(page).not_to have_content('CJ')
-      expect(page).to have_content('Sierra')
-      expect(page).not_to have_content('Adam')
-      expect(page).not_to have_content('Priscilla')
-      expect(page).not_to have_content('Indy')
-      expect(page).not_to have_content('Oscar')
-      expect(page).not_to have_content('Leona')
+      check "js-people-onLeave-checkbox"
```
kgf, line 53: Really nice. 
Could you do `expect(page_.to_not have_content('Leona')` right before the `check` ?

app/assets/javascripts/datatables_defaults.js
goldenhub77: @kgf This creates a div with class event-table-info specifically for adding in elements in JS to the datatables. This is the original revision which I plan on creating another commit to change but wanted you verify the functionality first

app/controllers/people_controller.rb
goldenhub77: @kgf I can change this in the commit to change the checkbox filter rendering.

app/assets/javascripts/datatables_defaults.js
kgf: Functionality looks really nice

spec/features/person_creation_spec.rb
goldenhub77: @kgf @ThomasCarlson Some of these forms are very large and have dynamic heights due to JS changing fields and content. This forces the page to scroll to bring the button in view. The arguments change depending on the height you want.

kgf:
Thanks again @goldenhub77 !!

Issue #539, opened 2017-10-09
Change item type to allow a tree structure


kgf:
On person show, there is a table to display items a person owns, rather than things the department owns. In that same accordion section, there is a button to create a New Item, but it doesn't work. This issue looks to either fix it or remove it. The challenge of fixing it is the hierarchy of items, so removing it and having people create new ones from the inventory page is acceptable. If you do chose to fix it, please include tests.
This is rated `easy` is you delete the button, `hard` if you fix it !

Thanks for your interest in this project,
Kevin

monoliths:
Hello, I can take this on! Its looks like there are two routes. Ill take the challenge for the harder solution :)

kgf:
Ah, challenge accepted, eh.
Thanks for your help @monoliths !

kgf:
@radvc This issue had already been claimed.
I am going to merge the PR from @radvc, since it does solve the broken button.
*However*, I'm not going to close the issue, as @monoliths is going to attempt to fix it, rather than just remove it. 
I heard what you said about the person#show being about a person, but these are only items the person owns, so still about the person.

monoliths:
Hey so I am looking for some feedback on a solution I had. Would following be more appropriate?  A new item dropdown where the dropdown consist of all the item types the new item can be where the item owner is set the person shown.

![screen shot 2017-10-16 at 12 32 06 am](https://user-images.githubusercontent.com/9068425/31600182-543d8ee4-b20a-11e7-9869-5bb4369493f3.png)

If not I can go the original route, just wanted to throw this idea out there.  Either way PR will come in sometime this week.

kgf:
I don't think this would be enough. There are currently two levels, and we're discussing adding the option of additional levels as needed. So it would need to step the user through the choices. So `Communications` might lead to `Portable Radio`, and then they could enter the specifics.

monoliths:
Ah ok. That makes sense.

kgf:
@monoliths Did you have other questions or ideas on this ?

monoliths:
Is the plan with "additional levels as needed" going forward or will this portion stick to two levels? 
Currently I've been looking at a small selection form that would sit below the items listing to start the process to create a new item. The form would use 2 selection fields, ItemCategory and ItemType that would use some coffee script/js to dynamically populate ItemType based upon what ItemCategory is selected. I can this working well it you are sticking with two levels. 

If this isnt clear I can share a screenshot once I get home.

[...]

PR #541, opened 2017-10-09
Remove new item button in accordion

radvc:
Reference for this issue https://github.com/ReadyResponder/ReadyResponder/issues/539. I remove the button because I think it is better if it only shows the information about the person or maybe we can add edit item functionality from that accordion.


kgf:
Thanks for your help, @radvc !

radvc:
@kgf Im always here. Thanks!

Issue #538, opened 2017-10-09
Events should show timecards.


kgf:
There are currently several tabs on an event, for things like tasks and availabilities. We would like to also have a tab for timecards, after notifications. It should show all timecards whose timespans overlap the event. The view should include the ability to verify them. This portion can be easily accomplished by using the table from the person#show view.

goldenhub77:
@kgf I can take this

kgf:
@goldenhub77 You're a machine !!

PR #543, opened 2017-10-10
events show timecards tab

goldenhub77:
Events should show timecards. #538


app/models/event.rb
```diff
@@ -132,6 +132,10 @@ def use_a_template
     end
   end
 
+  def timecards
```
kgf: I would prefer to have this in the Timecard model as a scope, where it can be reused.
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/models/availability.rb#L14
Also, I worry that people will see `@event.timecards` and think that it is a `has_many/belongs_to` relationship, which it is not.
Finally, you'll need to filter out cancelled timecards.

goldenhub77: @kgf I was thinking about a scope initially, I will attempt this again using that methodology and fix the logic as well.

app/views/timecards/show.html.erb
```diff
@@ -35,7 +35,7 @@
   %>
 <% end %>
 
-<table class="datatable table table-striped table-bordered" >
+<table class="generic-datatable table table-striped table-bordered" >
```
kgf, line 38: I thought this `generic` was what caused the heading issue in another table?

spec/factories/timecard.rb
```diff
@@ -1,5 +1,6 @@
 FactoryGirl.define do
   factory :timecard do
+    person
     start_time "2013-12-23 03:31"
```
kgf: Please do not add associations into the factory. I've found it to cause problems, because you don't control what gets created. What if person creates an assignment in it's factory, and assignments create a requirement, and a requirement creates a task, etc.
Instead, I prefer to create what I need, or exercise the option of stunning it out.

goldenhub77: @kgf I changed this back and created a person within the test separately.

app/views/timecards/show.html.erb
goldenhub77: @kgf I am not aware of a heading issues within another table, do you know which one was having issues? I'm pretty sure I resolved all of the heading issues.

kgf:
@goldenhub77 Thanks for your continued contributions !

Issue #544, opened 2017-10-10
Create a rack task to handle stale timecards


kgf:
People sometimes forget to sign out. Then the remember a few days from now, and sign out then. This results in a timecard with 100 hours. While I would also like validation on timecards, this is a request for a rake task to put any incomplete timecards into an error state after 24 hours. Please do not put much logic in the rake task itself, as they are difficult to cover in tests. 

For bonus points, pull the number of hours from a setting in the database. This would also want a migration to create the setting, and an addition to seeds.rb.

nesteves:
Hello, I would like to help with this.

For the migration, we're not going to change the database structure here, just add a new setting right? Something like:

```ruby
Setting.create name: 'Timecard hour limit', description: '..', key: '..', value: 24, 
```

Can you also elaborate on the timecard validation?

kgf:
Great, you're going for the bonus!
The setting should be retrieved from the database using the key only. Something like `timecard_maximum_hours` maybe...
Thanks for your help

vcavallo:
Adding myself here as a backup contributor in case @nesteves drops out or would like a hand 👋

kgf:
@vcavallo Could I interest you in a different issue ? Is there an area of software development that interests you ? There are additional improvements to RR that are not yet documented. I am happy to create a issue tailored to your interests. 
We appreciate your interest.
Kevin

vcavallo:
@kgf very cool!
Most of my experience is in Rails and TDD/BDD. I'm entirely competent on the front end, but typically prefer back end. Happy to take a gander at whatever you've got ;)

PR #548, opened 2017-10-11
Rake task to mark stale timecards with an Error status

nesteves:
As requested in issue #544 , this pull request implements the rake task `timecards:mark_stale_as_error` that looks for a `Setting` with the key `TIMECARD_MAXIMUM_HOURS` and marks all Timecards that have been open for more than *setting_value* hours with `status: 'Error'`. The `TIMECARD_MAXIMUM_HOURS` setting can be added to the `settings` table via the application interface (I also put an entry in `db/seeds.rb` that does that), or as an ENVironment variable.

I also included more specs for the timecard model and refactored it a bit.
**NOTE**: found a bug on the `most_recent` scope, which was listing Timecards starting with the ones with the oldest `start_time` and fixed it.


kgf:
@nesteves Thanks for all your work so far. 
Looks like there is a small amount of clean-up needed. The test that failed is in timecards spec.

nesteves:
My bad, I am having some trouble running the feature specs on a VM sitting behind a proxy and unwittingly caused a regression when I changed the timecard factory.

kgf:
No worries, that's what CI is for !

spec/models/timecard_spec.rb
```diff
@@ -19,60 +55,182 @@
...
-      expect(timecard).to be_valid
+  describe 'older_than scope' do
+    it 'returns a chainable relation' do
+      expect(described_class.older_than(1)).to be_a_kind_of(ActiveRecord::Relation)
     end
+    
+    context 'given 2 timecards with different start_times' do
+      before(:example) do
+        @timecard_with_7hours = create(:timecard, person: @cj,
+          start_time: 7.hours.ago, end_time: 1.hour.ago)
```
kgf: I really like all the tests!
But wouldn't this be a timecard with six hours ? ( 7-1)

spec/models/timecard_spec.rb
```diff
@@ -19,60 +55,182 @@
...
-      timecard = build(:timecard, person: nil)
-      expect(timecard).not_to be_valid
+      it 'returns timecards with a start_time older than 3 hours' do
+        expect(described_class.older_than(3)).to contain_exactly(
+          @timecard_with_7hours)
+      end
+
+      it 'returns timecards with a start_time older than 1 hour' do
+        expect(described_class.older_than(1)).to contain_exactly(
+          @timecard_with_7hours, @timecard_with_2hours)
```
kgf: This makes the duration more important, as the shorter card is actually one hour. This could lead to an intermittent test, as it's right at the line. Maybe change the end times to `30.minutes.ago` ?

kgf:
Nice work. A tiny change requested.
Happily, I learned some things as well!

spec/models/timecard_spec.rb
nesteves: Heh, I mean to describe the "age" of the start_time attribute, I'll change these to x_hours_old_timecard. Edit: oops, starting variable names with number is not allowed ofc..

kgf: Right, but I would like to have a wider margin for the assertion too.

spec/models/timecard_spec.rb
```diff
@@ -19,60 +55,181 @@
...
-      timecard = build(:timecard, person: nil)
-      expect(timecard).not_to be_valid
+      it 'returns timecards with a start_time older than 3 hours' do
+        expect(described_class.older_than(3)).to contain_exactly(
+          @timecard_started_7hrs_ago)
+      end
+
+      it 'returns timecards with a start_time older than 1 hour' do
+        expect(described_class.older_than(1)).to contain_exactly(
+          @timecard_started_7hrs_ago, @timecard_started_2hrs_ago)
```
kgf, line 183: Perfect, wide margin...

kgf:
Thanks for all your help, @nesteves !

PR #549, opened 2017-10-12
readjust width of datatable to fix accordion bug

goldenhub77:


kgf:
@goldenhub77 codacy reports issues that appear to be js related.

goldenhub77:
@kgf This solution came directly from data tables documentation as a valid solution for this issue. Maybe someone can find a better solution.

PR #553, opened 2017-10-13
fixed people.js.coffee google API and zipcode validation, moved people referrer methods to application_controller

goldenhub77:


app/models/person.rb
```diff
@@ -203,6 +203,11 @@ def service_duration
 
   private
 
+  def check_zipcode
+    return unless zipcode.present? && !zipcode.match(/^(?:[1-9]|0(?!0{4}))\d{4}(?:[-\s]\d{4})?$/)
+    errors.add(:zipcode, "invalid format - must be 5 digits, ex. 12345")
```
kgf: Should we change this error message to match the regex ? Zip and zip+4 `02134-1234` are both valid in the regex.

spec/models/person_spec.rb
```diff
@@ -23,6 +23,15 @@
       end
     end
 
+    it "zipcode must be 5 digits" do
```
kgf, line 26: Could we add one happy path assertion here ? `expect(correct_person).to be_valid`

kgf:
Sorry for the delay in review. Two small changes, then I think we're set here.

Thanks for your help !

PR #559, opened 2017-10-14
Upgrade from rails 4 to rails 5

sarat4a4:


kgf:
Hello @sarat4a4 Thanks for your interest in our project. However, this PR is not helpful. Were not ready for it yet. Perhaps you could take a look at our issues list and find something there that would interest you?

PR #561, opened 2017-10-14
Alphabetically sorting names in Dropdown, fixes #556

### LINKED ISSUE #556: When creating a new assignment, names in drop down are random.
It would be better to sort them, so you can find who you looking for.

umanghome:
Fixes #556.

Issue: While creating a new assignment, the names were random and not alphabetically sorted. [I call this Issue 1 for the screenshots.]

Another issue I encountered: While creating a new availability, the names were random, and an extra (empty) field in the dropdown was being created. [I call this Issue 2 for the screenshots.]

This PR fixes both these issues.

What I created to reproduce the issue:
* Added two records under "People": `Abe Lincoln` and `Umang Galaiya`
* Created a new scheduled Event, a Task under that event, and a Requirement under that task.

---

### Screenshots

Issue 1, before the fix:

image

Issue 1, fixed

image

Issue 2, before the fix

image

Issue 2, fixed

image

I will revert the fix for Issue 2 if I broke the intended behavior.

### DIFF
--- a/app/controllers/availabilities_controller.rb
+++ b/app/controllers/availabilities_controller.rb
@@ -30,6 +30,7 @@ class AvailabilitiesController < ApplicationController
     
     # Ensure person is included in @people_collection
     @people_collection |= [@availability.person] if @availability.person
+    @people_collection = @people_collection.sort_by {|firstname, middleinitial, lastname| "#{firstname} #{middleinitial} #{lastname}"}.compact
   end
 
   def create
--- a/app/controllers/availabilities_controller.rb
+++ b/app/controllers/availabilities_controller.rb
@@ -20,8 +20,9 @@ class AvailabilitiesController < ApplicationController
                                      end_time: event.end_time)
     @availability.person = person if person
 
-    @people_collection = Person.active
-    @people_collection |= [person]
+    people = Person.active
+    people |= [person]
+    @people_collection = people.sort_by {|firstname, middleinitial, lastname| "#{firstname} #{middleinitial} #{lastname}"}.compact
   end
 
   def edit
--- a/app/views/assignments/_form.html.erb
+++ b/app/views/assignments/_form.html.erb
@@ -3,7 +3,7 @@
 
   <div class="form-inputs">
     <%= f.association :person,
-                collection: @people_collection %>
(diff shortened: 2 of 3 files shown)


app/controllers/availabilities_controller.rb
```diff
@@ -20,15 +20,17 @@ def new
...
 
-    @people_collection = Person.active
-    @people_collection |= [person]
+    people = Person.active
+    people |= [person]
+    @people_collection = people.sort_by {|firstname, middleinitial, lastname| "#{firstname} #{middleinitial} #{lastname}"}.compact
   end
 
   def edit
     @people_collection = Person.active
```
kgf, line 30: Given that you are re-assigning it on line 25, should we drop this line ?

app/views/assignments/_form.html.erb
```diff
@@ -3,7 +3,7 @@
 
   <div class="form-inputs">
     <%= f.association :person,
-                collection: Person.active.where(:department => @requirement.event.departments) %>
+                collection: Person.active.where(:department => @requirement.event.departments).sort_by {|firstname, middleinitial, lastname| "#{firstname} #{middleinitial} #{lastname}"}.compact %>
```
kgf, line 6: Not needed for this PR, but that's a lot of logic to have in the controller. We should move it in a later PR.

umanghome: I'm sorry, I thought this issue is what this PR is supposed to fix. Am I missing something?

app/controllers/availabilities_controller.rb
umanghome: Which line are you suggesting we drop?

app/views/assignments/_form.html.erb
kgf: Disregard that comment

app/controllers/availabilities_controller.rb
kgf: All the numbers I'm using are the new line numbers. 
I suggest we get rid of line 29, as you re-assign it in line 33. The assignment at line 29 never gets used.

kgf:
@umanghome Thanks for your help !

app/controllers/availabilities_controller.rb
umanghome: It does get used, though. I'm using the assignment from line 29 and then sorting it.

umanghome:
@kgf No problem. :)

This was my first experience with Ruby.

PR #562, opened 2017-10-14
#560 Display dates & editor on availability

### LINKED ISSUE #560: Display dates and editor on availability
On many show pages, there is a footer that shows the created_at, edited_at and last_editor. This is not present on availability, but we'd like it to be.
This is an example from a different page
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/views/item_categories/show.html.erb#L17
Any questions feel free  to ask.

Atul9:
closes #560


kgf:
@Atul9 Thanks for your contribution ! We appreciate your help.

Issue #343, opened 2017-01-31
Validation needed on id_code for notification


kgf:
Both Events and Notifications have id_codes. The validation would need to check both for duplicates. 
Id_code is optional for a notification, but if present, must not be a duplicate.
Also, an id_code can be recycled once the event/notification is cancelled/complete/closed. Even then, to prevent confusion, they should probably wait six months or so before being valid again.

rhemaboyo:
Hi, can I go ahead and work on this issue?

kgf:
@rhemaboyo That would be great, thanks !
Post here if you have any questions.
We appreciate the help.

rhemaboyo:
For the event model I was thinking about validating uniqueness unless the most recent event with that id_code ended over 6 months ago. 

```
validates_uniqueness_of :id_code, :unless => :expired?
...
def expired?
    # 1.577e+7 = ~ 6months
    Event.where(:id_code => id_code).maximum('end_time') - Time.now + 1.577e+7 < 0
end
```

I was wondering if you want the notification model to have its own id_code or does it just share the id_code of the event it belongs to? (In which case the above validation would apply to both.)

kgf:
A notification either has its own unique id code, or it has none.
Rails includes time helpers, so you can say `6.months.ago`. That would read better, I think.

kgf:
@rhemaboyo We are quite close on this, hoping you'll be able to resolve the failing test.

kgf:
Closed by #593

PR #564, opened 2017-10-14
Uniq id code

rhemaboyo:
#343 added id code column to notifications table and added validation for uniqueness. For both events and notifications, made sure same id_code could be added if event ended >= 6 months ago.


app/models/event.rb
```diff
@@ -3,7 +3,8 @@ class Event < ActiveRecord::Base
   before_save :calc_duration, :trim_id_code
 
   validates_presence_of :category, :title, :status, :id_code
-  validates_uniqueness_of :id_code, :title
+  validates_uniqueness_of :title
+  validates_uniqueness_of :id_code, unless: :expired?
```
kgf, line 7: This seems awkward. I would not expect to do uniqueness on a method, especially one ending in `?`

app/models/event.rb
```diff
@@ -142,4 +143,9 @@ def calc_duration #This is also used in timecards; it should be extracted out
   def trim_id_code
     self.id_code = self.id_code.split[0].downcase
   end
+
+  def expired?
```
kgf, line 148: I would expect a method ending in `?` to return a boolean.

app/models/event.rb
rhemaboyo: it's validating the uniqueness of the id_code attribute, not the method, unless it is expired - meaning the one in the database is over 6 months old. Did you not want to check for how old the one in store is?

app/models/event.rb
rhemaboyo: Yes, it will return either true or false. If True it will not validate uniqueness, if false it will.

app/models/notification.rb
```diff
@@ -71,4 +72,9 @@ def notification_has_at_least_one_recipient
     end
   end
 
+  def expired?
+    Notification.where(id_code: id_code)
+      .select { |notification| notification.end_time > 6.months.ago }.empty?
```
kgf, line 77: Ah, now this one makes more sense, because it has the `.empty?` on the end.

app/models/event.rb
kgf: I'm sorry, I completely read this wrong, and I don't quite know why ! I read the `unless` as another attribute being checked, as if it were a list of three things being checked for uniqueness.
Never mind, moving on.

app/models/event.rb
kgf: I think I like the style you used in the notification model. Won't this return either `nil` or a collection of Events ? Admittedly that does end up being either truthy or falsey. 

The `empty?` helps.
Thanks !

rhemaboyo: I'll go ahead and update that. It makes sense for the methods to match anyway.

rhemaboyo: Done.

kgf:
@rhemaboyo I'm ready to merge the changes as proposed, but sadly the tests did not pass. 

Thanks for your help.
Kevin

kgf:
I took a quick look at the failed tests. One is about the id_code being already taken. If we changed the factory to include `sequence`, I think that might resolve it.
`sequence(:id_code)`

kgf:
@goldenhub77 may have fixed this test. I just rebased, we'll see

rhemaboyo:
Hey @kgf , so sorry for the late response. Would you be able to paste a snippet of the failed test so that I can have an idea of what is going wrong? I'm not exactly sure what `sequence(:id_code)` does...

kgf:
No problem @rhemaboyo , we all get busy.
If you take a look below, you'll see the line for travis. The end of the line has a link for `details`. If you  click on that link, you will be taken to Travis, where you can get all the details.
Let me know if you have any other questions.
Thanks for your help. 
Kevin
screen shot 2017-10-19 at 8 07 23 pm

kgf:
Regarding `sequence(:id_code)` that is a line for Factory Girl. It keeps track of which numbers have already been used, and uses the next in the sequence. Avoids problems when you have validations for uniqueness.

app/models/event.rb
```diff
@@ -143,4 +144,10 @@ def calc_duration #This is also used in timecards; it should be extracted out
...
+    Event.where(id_code: id_code)
+      .select { |event| event.end_time > 6.months.ago }.empty?
```
pcai, line 150: Not sure if there's a constraint I'm not aware of, but may I suggest this instead? `Event.where(id_code: id_code).where("end_time > ?", 6.months.ago).empty?`

Chaining the `where` clause allows the database to return yes or no, saving us the work of constructing the objects in ruby, searching through them, only to immediately count and discard them.

pcai: Also, if the event is already persisted, this search needs to exclude `self.id`, or the record may appear to conflict with itself.

app/models/notification.rb
pcai: Same here - if the notification is already persisted, I think this search needs to exclude `self.id`, or the record may appear to conflict with itself during an update.

pcai:
Thanks for working this issue. Right now, I think adding a uniqueness validation on notifications and events causes a full table scan on every save. We can probably resolve this by adding an index on `id_code` on both tables, since we'll be searching by that column frequently.

The travis failures could be fixed by adding a sequence to the factory girl definitions, but we might be able to fix some of them by only validating uniqueness if the `id_code` is present. I think that's the desired behavior if I'm reading https://github.com/ReadyResponder/ReadyResponder/issues/343 correctly. Right now, `id_code` isn't explicitly validated for presence, so it may have the effect of allowing a nil value, but then enforces uniqueness on it every 6 months.

kgf:
@pcai is correct, a notification does not need an id code. Only events are required to have one.

Issue #272, opened 2016-10-28
Add comments as a distinct model to Person


kgf:
I am scoping this to only person to make multiple small issues.
Comments will be used on many models, such as items, tasks, pretty much everything. Most things will be `commentable`. I think it would be good if this was implemented as a Rails concern, so it could be easily added to additional models going forward. I am open to suggestions in that regard. There should be a table at the bottom of person showing all comments, most recent at the top. That enough for this issue.

https://richonrails.com/articles/rails-4-code-concerns-in-active-record-models

abhishekbose87:
I would like to take a jab at this one.

kgf:
Excellent, thanks for your help.
Kevin

kgf:
@abhishekbose87 did you have any success? Do you have questions?

kgf:
This work has been started, as can be seen here.
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/models/comment.rb

alexshook:
@kgf hey do you still need someone to do this?

kgf:
Absolutely, @alexshook that would be great. Thanks !

Any questions ?

alexshook:
Not right now, @kgf, but will let you know if I think of any. Thanks!

kgf:
@alexshook Any updates ? Are you stuck on anything ?

alexshook:
@kgf I have a branch with changes that I'd like your feedback on before proceeding with any further. Has become a busy day so I'll open a PR for you with some questions tomorrow.

kgf:
@alexshook That's great ! I look forward to seeing the PR tomorrow.

kgf:
Closed by #565 , although a follow-up issue will be needed to create the comments.

PR #565, opened 2017-10-15
Add comments to persons

alexshook:
For #272. The original description doesn't say anything about creating comments, only showing them, so I left out create for now but am happy to do it if you give me a few requirements. I added a table to the people show view to display comments. Should this be added anywhere else?


spec/models/comment_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe Comment do
+  describe 'validations' do
+    it { is_expected.to validate_presence_of :person_id }
```
alexshook: Added some validations before I realized I didn't know what to do about `create`, can remove them if you'd rather have `create` be part of a different story.

app/controllers/people_controller.rb
```diff
@@ -73,7 +73,7 @@ def index
   end
 
   def show
-    @person = Person.find(params[:id])
+    @person = Person.includes(:comments).find(params[:id])
```
alexshook, line 44: I waffled a bit between what I did here and going for `@comments = @person.comments`, similar to `@phones` on line 79. Let me know if you have a preference.

kgf: This has the advantage of fewer sql queries.

app/models/comment.rb
```diff
@@ -1,3 +1,7 @@
 class Comment < ActiveRecord::Base
   belongs_to :commentable, :polymorphic => true
+
+  scope :recent, ->{ order('created_at DESC') }
```
kgf: This scope confuses me. It is named `recent`, but it does not limit the comments based on time. Perhaps name it `by_creation_date` or something like that.

kgf: Thinking about it more, you could even make this a default scope, because we would rarely want the unsorted.

alexshook: 👍 sounds good to mee

kgf:
Sorry I missed your question earlier. Yes, it would be nice to be able to enter them too ! 
Are you doing OK other than that ?

app/models/comment.rb
```diff
@@ -1,3 +1,7 @@
...
+
+  validates_presence_of :person_id, :description
```
kgf: You might not have noticed, but this is a polymorphic class. See line 2 of this file. I don't think we can validate on person_id.

alexshook: I believe `person_id` will validate (we should always have the concept of who posted the comment?) since that's stored in the comments table itself. At least I can't think of any reason why it wouldn't, but also haven't touched a polymorphic association in a while. 

I did think that `person_id` might better be called `poster_id`. That might be confusing for when the commentable type and id are a Person though, so... idk. I'll remove the validation if it ends up being a 💣 while implementing create comments.

alexshook:
Yes I'm totally good with creating comments as well. Could you give me some
guidance on what you're looking for? Something like where to put a text
field for comments, if you'd like any specific labels for the
fields/placeholder text, etc.? I mostly make APIs so I'm not the best at
deciding where front end elements should go.

kgf:
Are you finding this work enjoyable? I could create an issue closer to API work for you.

We can almost merged this here, although it won't quite work. The validation on person id won't work,  as comments are polymorphic.

alexshook:
Yeah this is great, always good to do slightly different type of work...
and also to not completely forget how to make a form. I'll remove the
person_id validation now, although I would swear that it would work.


*Alex Shook610.703.8914 | alex.e.shook@gmail.com <alex.e.shook@gmail.com>*
*LinkedIn <http://www.linkedin.com/in/alexshook> | @alexshook
<https://twitter.com/alexshook> | Portfolio <http://alexshook.com/>*

kgf:
The person validation might work for this PR, but we're going to add comments to items and events as well. 
Thanks for your help.
Kevin

kgf:
Sadly, the tests did not pass...

alexshook:
It looks like the certs tests failed, can you rerun the build?

kgf:
@alexshook build restarted

kgf:
@alexshook thanks for your help. 
This partially addresses #272 .

PR #566, opened 2017-10-15
fix date and editor

abhinavm27:


kgf:
@abhinavm27 Thanks for your interest in the project. 
This is too large a PR to accept. It also does not pass the tests. You may notice that you repeated the same code multiple times. My intent, after we get more of the views covered, is to move this to a shared view, such as the footer.
I hope you'll take a look at the list of issues, and choose to tackle one of those.

Kevin

Issue #537, opened 2017-10-09
Entering a repair should update the item


kgf:
When entering a repair, you can describe the repair, but that's it. This change would add a `Condition` field to the Repair model. It should have the same choices as `Condition` an Item has. When you save a repair, it should set the condition of the item to the same condition as indicated on the repair.

This will need further refinement in the future, because an item may need  more than one repair at a time. For now, this will improve the UX.

dusty-wil:
@kgf I'm interested in looking in to this! Does this mean that the repair database table is going to need a new field also, or should the repair update the item only?

kgf:
Hmm, good question. My instinct is that we'll want that extra information. ( How did this item get to the condition.) Let's still call it `condition`.

dusty-wil:
@kgf sorry for the delay on this... I had to step back to deal with some school stuff this week. If this is time sensitive and you want to assign it to someone else please feel free, otherwise I'll likely be able to look at it this weekend!

kgf:
NP, things happen. This weekend will be fine. Thanks for your help! And also thanks for the update !

dusty-wil:
@kgf Ok, so I've created the table field, and when I save a repair I can see the condition is saved with the repair. However, I'm not sure how to update the repair status for the item. Should I be accessing the item's properties through the repair's controller? Is there another spot where this kind of update happens so I can get a general idea of how that should work?

kgf:
I would probably approach it in the Repair controller. You'll probably want to create a method in the Item to calculate it's condition, which would get called whenever a repair is created or updated. This method would find all it's repairs that did not have a condition of `Ready`, then set the item to the worst of those conditions.
I hope that's clear, feel free to ask questions.
Thanks for your help,
Kevin

dusty-wil:
Thanks @kgf, that is clear. I've got the method created and it can update the item's condition, but I'm wondering, should the method instead update the condition of the item to the most recent service date? IE if an item's in "In Service - Maintenance" I can add a repair that changes the item's state to "Ready" ... or maybe I'm not understanding how the repair should work?

kgf:
Repair would track one issue from start to finish. 
Say I have a truck. I create a repair to have the inspection sticker updated. It would have a condition of`In service-Degraded` I then make a second repair, flat tire, `Out of service`. When the tire gets fixed, the condition of that repair is`ready`. be But the item condition should go back to degraded. That repair issue is still open.

dusty-wil:
[...]

PR #568, opened 2017-10-18
 #537 - Entering a repair should update the item

dusty-wil:
Addresses issue #537. Added field for condition in repair table, created functionality to update item's condition based on condition chosen in repair. Added condition column to repair display tables and created a condition field when adding/editing a repair.


app/models/item.rb
```diff
@@ -51,4 +52,13 @@ def location_name
     "Unknown"
   end
 
+  def repair_condition
+    severity_index = 0
+    repairs.each do |a|
```
kgf: Rather than iterating over ALL the repairs, perhaps we should create a scope that would only show the ones where the condition is not `Ready`. A similar example is here
https://github.com/dusty-wil/ReadyResponder/blob/development/app/models/repair.rb#L24

app/models/item.rb
```diff
@@ -51,4 +52,13 @@ def location_name
     "Unknown"
   end
 
+  def repair_condition
```
kgf: I would like this to be an active verb, more descriptive. Imagine yourself seeing it in 6 months. What does it do ?
Perhaps `set_repair_condition` ?

app/views/repairs/index.html.erb
```diff
@@ -12,6 +12,7 @@
       <th>Service date</th>
       <th>Status</th>
       <th>Description</th>
+      <th>Condition</th>
```
kgf, line 16: Since you're already altering this table, please add the item name.

kgf:
Some comments made.

dusty-wil:
Ok, sounds good! I can look at making those changes!

kgf:
Sadly, the tests failed.

dusty-wil:
Ok, I've made those changes. I see the check is still failing... is this because I need to add db/schema.rb to my commit as well?

It looks like this is the error the check is reporting:

bundler: failed to load command: rspec (/home/travis/build/ReadyResponder/ReadyResponder/vendor/bundle/ruby/2.3.0/bin/rspec)

ActiveRecord::PendingMigrationError: 

Migrations are pending. To resolve this issue, run:

	bin/rake db:migrate RAILS_ENV=test

dusty-wil:
Yep, that seems to have been the issue

kgf:
@dusty-wil Thanks for your help !

dusty-wil:
@kgf no problem, thanks for the guidance! I definitely feel like I'm starting to understand how a ruby/rails application works

kgf:
We'll, we're happy to mentor as well, if you're willing to continue to help.
Thanks for this improvement.
Kevin

dusty-wil:
@kgf thanks! Things got busy with school for a bit, but I'm planning on looking for more issues to tackle next week.

kgf:
Great ! We can use **all** the help we can get !

PR #569, opened 2017-10-18
Minimum Title! 

andyvh:
Added the field min_title (denoting the lowest rank required to attend) to events! Now, "upcoming" should only yield messages for those of sufficient rank or higher!


kgf:
@andyvh You'll need to rebase

app/models/event.rb
```diff
@@ -88,7 +88,7 @@ def responding_people
   end
 
   def eligible_people
-    Person.active.where(department: self.departments)
+    Person.active.where(department: self.departments).where("title_order >= ?", Person::TITLE_ORDER[min_title])
```
kgf, line 93: This is nice, all the results with only a single call to the database.

db/migrate/20171009043433_change_min_rank_type.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeMinRankType < ActiveRecord::Migration
+  def change
+    change_column :events, :min_title, :string, :default => Person::TITLES[Person::TITLES.length - 1]
```
kgf, line 3: I would rather not have this default here. This would be set when the migration runs. If the list of Titles changes later, you could have a default that wasn't in the list. This is more likely in the future, as we move the titles into the department model.

spec/models/event_spec.rb
```diff
@@ -34,6 +34,12 @@
     expect(@event.id_code).to eq("howdy")
   end
 
+  it "only sends notifications to above a certain rank" do
```
kgf, line 38: A test, excellent

db/migrate/20171009043433_change_min_rank_type.rb
andyvh: Understood! I'll fix it as soon as I've rebased! It shouldn't be long, but I'm a little new to this, so I'm moving tentatively...

andyvh: OK, I've removed the default!

PR #570, opened 2017-10-19
 implementation of calendar view on events index

goldenhub77:


kgf:
@goldenhub77 once again, thanks for your help.

PR #571, opened 2017-10-19
[RFR]created vendor table

jacklynhma:
Migrated the source data to another column, source_data, just in case we needed to retrieve the information. Created a vendors table, form and show page. Added a button to add vendors in the new items form, which will redirect to the new items form with the help of the params[:next] 

Created controller and model tests.


kgf:
This is a start, you're on the way. Feel free to ask questions as needed.

db/migrate/20171019014656_create_vendor.rb
```diff
@@ -0,0 +1,13 @@
+class CreateVendor < ActiveRecord::Migration
+  def change
+    create_table :vendors do |t|
+      t.string :name
+      t.string :street
+      t.string :city
+      t.string :state
+      t.string :zipcode
+      t.string :status
+      t.text :comments
```
pcai: missing `t.timestamps` - would that be useful?

kgf: Hey it's @pcai ! Cool
timestamps are a good idea.

db/schema.rb
```diff
@@ -268,6 +268,7 @@
     t.string   "condition"
     t.integer  "qty"
     t.integer  "grant_id"
+    t.integer  "vendor_id"
```
pcai: needs an index please

app/views/vendors/_form.html.erb
```diff
@@ -0,0 +1,15 @@
+<h1> Add a Vendor </h1>
+<%= simple_form_for @vendor do |f| %>
+<%= f.error_notification %>
+
+  <%= hidden_field_tag :next, params[:next] %>
```
kgf: What does this line do ?

jacklynhma: When the new vendor form is submitted, this gives it the ability to redirect to the new items page.

db/schema.rb
jacklynhma: will add it!

db/migrate/20171023011304_migrate_data_to_vendor.rb
```diff
@@ -0,0 +1,18 @@
+class MigrateDataToVendor < ActiveRecord::Migration
+  def up
+
+    Item.all.each do |item|
+      v = Vendor.where(
+        name: item.source
+      ).first_or_create
+
+      item.vendor_id = v.id
```
pcai, line 9: I think you need to save the `item` here so the change persists?

jacklynhma: ok!

pcai: Sorry I should've been more clear: try something like `item.update_columns(vendor_id: v.id)` to skip callbacks and validations, in addition to being faster it is likely to be safer than:

```ruby
item.vendor_id = v.id
item.save
```

in this example, if a validation fails it will silently discard the update which we probably don't want; once the `source` column is removed we permanently lose the data

app/controllers/vendors_controller.rb
```diff
@@ -0,0 +1,30 @@
+class VendorsController < ApplicationController
```
pcai, line 1: I think codacy is complaining because there's no authorization check on this controller - see items controller for example, it will have these lines:

```
  before_filter :authenticate_user!
  load_and_authorize_resource
```

need to update ability.rb to give existing roles the ability to edit vendors.

ref: https://github.com/CanCanCommunity/cancancan/wiki/authorizing-controller-actions

jacklynhma: done!

kgf:
@jacklynhma thanks for your continued efforts !

kgf:
I think you'll need to remove this line, for a start
https://github.com/ReadyResponder/ReadyResponder/blob/development/spec/factories/item.rb#L7

app/controllers/vendors_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+  before_filter :authenticate_user!
+  load_and_authorize_resource
+
+  def index
+    @vendors = Vendor.all
+  end
+
+  def show
+    @vendor = Vendor.find(params[:id])
```
pcai: this may result in double loading since `load_and_authorize_resource` loads and authorizes `Vendor.find(params[:id])` and assigns it to `@vendor`. I think you can remove it and the tests will continue to pass?

app/controllers/vendors_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    @vendor = Vendor.new
+  end
+
+  def create
+    @vendor = Vendor.new(vendor_params)
+
+    if @vendor.save
+      redirect_to params[:next] || vendors_path, notice: "Vendor was successfully created"
+    else
+      render action: new
```
pcai: nitpick: when this fails, I think the validation errors are not set in the flash?

app/views/vendors/index.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    <tr>
+      <th>List of Vendors</th>
+    </tr>
+  </thead>
+
+  <tbody>
+
+    <% Vendor.all.each do |vendor| %>
+      <tr>
+        <td><%= link_to "#{vendor.name}", vendor_path(vendor.id)%></td>
```
pcai: also known as `<%= link_to vendor.name, vendor %>`

app/views/vendors/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  <thead>
+    <tr>
+      <th>List of <%= @vendor.name %> Items</th>
+    </tr>
+  </thead>
+
+  <tbody>
+    <% @items.each do |item| %>
+      <tr>
+        <td><%= item.name%></td>
```
pcai: hmmmm it would be more useful if these were clickable links to the items, right?

app/views/vendors/index.html.erb
```diff
@@ -0,0 +1,22 @@
...
+
+  <thead>
+    <tr>
+      <th>List of Vendors</th>
+    </tr>
+  </thead>
+
+  <tbody>
+
+    <% Vendor.all.each do |vendor| %>
```
pcai: in the controller you assign `@vendors = Vendor.all` but then you don't use `@vendors` anywhere

app/views/layouts/_top_menu.html.erb
```diff
@@ -41,6 +41,10 @@
             <%= nav_link "Inventory", item_categories_path if can? :read, Item %>
           </li>
 
+          <li>
+            <%= nav_link "Vendors", vendors_path if can? :read, Item %>
```
pcai: typo: should be `can? :read, Vendor`, (not `Item`)

kgf:
@jacklynhma Thanks for your continued efforts !
And thanks to @pcai for the review !

app/controllers/vendors_controller.rb
jacklynhma: yep the tests still pass

jacklynhma:
ready for review!

kgf:
Thanks for your help, @jacklynhma. 
Sorry for the delay, I've been at the AWS conference. I'll take a look soon.

app/views/layouts/_top_menu.html.erb
```diff
@@ -41,6 +41,10 @@
             <%= nav_link "Inventory", item_categories_path if can? :read, Item %>
           </li>
 
+          <li>
```
kgf: Could you please move this under the inventory menu item ? I don't think it warrants a top level spot. 
See Inactive, under People for an example.
Keep the permissions check, that's good.

app/views/items/_form.html.erb
```diff
@@ -1,7 +1,12 @@
 <%= simple_form_for @item, :html => {:multipart => true} do |f| %>
+
   <%= f.error_notification %>
 
   <div class="form-inputs">
+    <%= f.association :vendor, label: "Vendor ", collection: Vendor.order("name") %>
+
+    <%= link_to "Add a New Vendor", new_vendor_path(next: request.path) %>
```
kgf: Could you please add a permissions check onto this ? Thanks

app/views/vendors/index.html.erb
```diff
@@ -0,0 +1,18 @@
+
+<table id="List_of_vendors" class="generic_datatable table table-striped table-bordered">
+  <caption><h3><%= @page_title %></h3></caption>
+
+  <thead>
+    <span>
```
kgf: What does this span do for us ?

kgf:
Sorry for the delay @jacklynhma , we do appreciate your help !
I made some comments, looking forward to your responses.

app/views/layouts/_top_menu.html.erb
jacklynhma: 👍

app/views/items/_form.html.erb
jacklynhma: 👍

app/views/vendors/index.html.erb
jacklynhma: It does nothing. deleted

jacklynhma:
@kgf ready for review! all tests passed locally

app/views/layouts/_top_menu.html.erb
```diff
@@ -18,6 +18,7 @@
             <ul class="split-dropdown dropdown-menu">
               <%= nav_link "Applicants", applicants_people_path %>
               <%= nav_link "Inactive", inactive_people_path %>
+              <%= nav_link "Vendors", vendors_path if can? :read, Vendor %>
```
kgf: This should not be under people, as it's not about the members. We'll need to create a submenu under inventory, similar to what we have now for people and events.

app/views/items/_form.html.erb
```diff
@@ -1,7 +1,12 @@
 <%= simple_form_for @item, :html => {:multipart => true} do |f| %>
+
   <%= f.error_notification %>
 
   <div class="form-inputs">
+    <%= f.association :vendor, label: "Vendor ", collection: Vendor.order("name") %>
```
kgf: This should not be the very first thing on the form. Probably leave it in the same position `Source` was in previously ?

spec/factories/item.rb
```diff
@@ -4,7 +4,7 @@
   factory :item do
     name "EMA-10"
     description "Ford F-150"
-    source "Main Street Ford"
+    association :vendor, factory: :vendor
```
kgf: I would prefer not to have the factory create associated objects. Doing this can lead to a cascade of creations. Any spec that needs it can create it, or stub it out.

spec/factories/vendor.rb
```diff
@@ -0,0 +1,11 @@
+FactoryGirl.define do
+  factory :vendor do
+    name "EMA-29"
+    street "Washington st "
+    city "nyc"
+    state "NY"
+    zipcode "10036"
+    status "Avaiable"
```
kgf: Typo, and double quotes

app/views/vendors/show.html.erb
```diff
@@ -0,0 +1,17 @@
+<table id="vendor_show_page" class="generic_datatable table table-striped table-bordered">
+  <caption><h1><%= @vendor.name %></h1></caption>
+
```
kgf: We also need Vendor Status

app/views/items/_form.html.erb
kgf: Also, it should limit the choices to Vendors who are active status.

kgf:
There does not seem to be a vendor edit form, only new. I think we'll need one. Also, the new form does not have status.
Thanks for your help !
Kevin

app/views/layouts/_top_menu.html.erb
jacklynhma: done!

spec/factories/item.rb
jacklynhma: I will remove it

spec/factories/vendor.rb
jacklynhma: do you prefer single quotes?

app/views/vendors/show.html.erb
jacklynhma: added

app/controllers/vendors_controller.rb
```diff
@@ -24,9 +24,24 @@ def create
...
 
+  def edit
+    @vendor = Vendor.find(params[:id])
+  end
+
+  def update
+    @vendor = Vendor.find(params[:id])
+
+    if @vendor.update_attributes(vendor_params)
+      redirect_to vendor_path(@vendor), notice: "Vendor was successfully created"
```
pcai: should say "updated"

app/controllers/vendors_controller.rb
```diff
@@ -24,9 +24,24 @@ def create
...
+    else
+      flash[:alert] = @venor.error.full_messages.join(", ")
```
pcai: typo here

app/models/item.rb
```diff
@@ -7,7 +7,7 @@ class Item < ApplicationRecord
   #                 :model, :brand, :name, :owner_id, :po_number,
   #                 :value, :grant, :purchase_amt, :purchase_date,
   #                 :sell_amt, :sell_date, :stock_number,
-  #                 :source, :status, :condition, :comments, :item_image,
+  #                 :vendor_id, :status, :condition, :comments, :item_image,
```
pcai, line 10: due to the rails5 merge this needs to be updated to allow `vendor_id` through strong params - see https://github.com/ReadyResponder/ReadyResponder/blob/development/app/controllers/items_controller.rb#L61

PR #573, opened 2017-10-20
Feature/generate qrcode for add contact

kurotyan514:
* STORY
https://github.com/ReadyResponder/ReadyResponder/issues/2

add QRCode on person show page, user can use their phone to scan and have them added to their contacts.

please review by commit thanks! and feel free to ask


app/services/vcard/generator.rb
```diff
@@ -0,0 +1,37 @@
...
+    private
+
+    def generate_phones
+      person.phones.each do |phone|
+        vcard.tel phone.content, type: phone.category
+      end
+    end
+
+    def gender_abbr
+      person.gender == 'Male' ? 'Mr.' : 'Mrs.'
```
kgf: I think we probably don't need the gender. In America, `Mrs` signifies a married woman only.

app/services/vcard/generator.rb
```diff
@@ -0,0 +1,37 @@
...
+      @person = person
+    end
+
+    def call
+      @vcard = VCardigan.create(version: VERSION)
+      vcard.name person.firstname, person.lastname, '', gender_abbr
+      vcard.fullname person.fullname
+      vcard.email person.email
+      vcard.bday person.date_of_birth.strftime("%Y%m%d")
+      vcard.title person.department.name
```
kgf: Could we do this as the `org` instead of the title?

app/views/people/show.html.erb
```diff
@@ -4,6 +4,7 @@
   <%= sidebar_button_link 'New Availability', new_person_availability_path(@person) if can? :create, Availability %>
   <%= sidebar_button_link 'New Timecard', new_person_timecard_path(@person) if can? :create, Timecard %>
   <br>
+  <%= raw @qr %>
   <%= image_tag @person.portrait_url if @person.portrait.file.present? %>
```
kgf: Please put the portrait above the barcode. Thanks

kgf:
This is very good! Some small changes requested.

app/services/vcard/generator.rb
```diff
@@ -0,0 +1,33 @@
...
+    def self.call(person)
+      new(person).call
+    end
+
+    def initialize(person)
+      @person = person
+    end
+
+    def call
+      @vcard = VCardigan.create(version: VERSION)
```
pcai, line 14: is this a typo? its declared `@vcard` here but referenced as `vcard` in the rest of the method

pcai: sorry, I just saw the `attr_reader :vcard` declaration. But I am not so sure about the value of an instance variable that is only ever used in the same invocation of the method that sets it?

kgf:
@kurotyan514 In order to fix the test suite, you need to add `date_of_birth 25.years.ago` to the file at `spec/factories/person.rb`
Thanks for your help,
Kevin

Issue #449, opened 2017-09-21
When creating a new availability, it should cancel others shadowed. 


kgf:
Currently, a volunteer can text in that they are available using the ID code for an event, and that will be recorded as an availability with the same start and end times as the event matching that ID code. Note that there is no parent-child relationship between the two. If the volunteer then sends in another text saying unavailable, the first Availability will be found and it's status changed to cancelled. The new one will be saved. Great.

But,
If the volunteer sends in Available, as in the first example, and then sends in `Unavailable custom` ( `custom` allows them to sign-up for any arbitrary times ) for a period greater than the first record , both records will exist and conflict with each other. This is confusing and bad. This issue requests that if the new `custom` availability completely covers one or more existing availabilities, the existing ones should all be set to cancelled before the new one is saved.

@danielstrimpel is this of interest, since you were just in this area of the code ?

ghost:
Yep, I'll take this.

danielstrimpel:
In trying to recreate the issue, I tried to create a new person on my local dev machine to have my phone number. However, it wouldn’t let me create a person on submit, saying ["Phones content is invalid”] even though all fields were filled in. I’d make an issue for this, but I want to make sure it’s not just happening on my local setup for some reason, and I haven't found where the production URL is yet (help with that please?).

Another possible issue: trying to edit an existing person to add phone info doesn’t work, because the section is missing from the form as in this screenshot:
![ready](https://user-images.githubusercontent.com/21265184/30814675-dac08b7e-a1de-11e7-940f-b74fc21b4e2d.jpeg)

Dunno if that’s sposed to be that way, or if its another issue.

ghost:
In trying to recreate the issue, I found a bunch of problems with the New Person form. 

- [ ] Complains about wrong phone info incorrectly and won't submit (so can't create new person with my phone number to test).
- [ ] Edit person has no fields for phone info ( so can't add my phone number to existing person)
- [ ] In phone info, label for number says “content” (should be "number")
- [ ] Status choice “invalid” seems strange, why would I want to enter an invalid number? Same thing with email.
- [ ] What are "old comments" for a new person?
- [ ] I can't enter text into the “titles” text box
- [ ] DOB field: calendar doesn’t pop up

[...]

PR #575, opened 2017-10-22
New availability cancels shadowed ones

nesteves:
Implements the functionality requested in issue #449 

A user of the application that has a registered phone number can send a text message to the application's number indicating their availability with one of the following formats:

```
available event01

unavailable event01

available custom 2017-12-21 10:00:00 2017-12-21 18:00:00

unavailable custom 2017-12-21 10:00:00 2017-12-21 18:00:00
```

Any one of these will generate at least one availability in the system, with a number of side-effects, described below:

* if the new availability *contains* any existing ones, cancel those, no matter their status
* if the new availability *overlaps* any existing ones, that is, covers a part of their duration
  * having different statuses, move the existing one's boundary so they become contiguous
  * having the same status, cancel the existing one and expand the new one so it covers both
* if the new availability *is contained* by an existing one
  * having different statuses, break the old availability into 2 and accommodate the new one
  * having the same status, return an error message (no availability created)

----
This pull request adds the following to the codebase:

* a series of scopes on the `availability` model that make it easier to do range comparisons
* a validation that prevents the creation of new availabilities that overlap any already existing ones that belong to the targeted person and happen to be active
* a suite of feature specs for receiving availability messages, covering most cases
* a service object for the creation of new availabilities that manipulates existing availabilites to "make room" for the new one


nesteves:
One thing I remembered, when deployed to production, if there are any active and overlapping (thus invalid) availabilities for a person, she won't be able to message her availability for any periods where these same availabilities take place, as the ensuing updates will fail on validation.

While this might be an issue at the beginning, it will basically fix itself by letting time go by, unless availabilities are defined well in advance.

Any thoughts?

kgf:
I haven't even looked at the code yet, and this is already an impressive PR. The write-up at the beginning is great, thanks.
The only overlap is when people put in for vacation. I'm not too concerned. We'll watch for it.

app/models/availability.rb
```diff
@@ -17,11 +17,32 @@ class Availability < ActiveRecord::Base
...
   scope :available, -> { where(status: "Available")}
   scope :unavailable, -> { where(status: "Unavailable")}
+
   scope :in_the_past, -> { where("start_time <= ?", Time.zone.now) }
-  scope :active, -> { where(status: ['Available', 'Unavailable']).joins(:person).where(people: {status: "Active"}) }
+
+  # These scopes tie the application to postgres, as they rely
+  # on its range data type implementation. The ranges are left_bound closed
+  # and right_bound open, as specified by the argument '[)'
+  # reference: https://www.postgresql.org/docs/9.5/static/rangetypes.html
```
kgf, line 29: Thanks, I like the comment to more documentation

kgf:
Thanks @nesteves That was a great PR ! Lots of tests, interesting approach, and I'm learning about the PG time stuff.

Thanks for your ghelp,
Kevin

nesteves:
Thank you! I learned a lot too. 

Regards,
Nuno

Issue #15, opened 2014-03-02
Expired certs still grant the skill


kgf:
Expired certs should still be shown, but they should not grant the skill.

kgf:
This would be good to have !

cesc1989:
@kgf would you mind explaining a bit more about this to me? I still struggle to understand some terminology specific to this project.

Thanks! :)

kgf:
@cesc1989 Thanks for your interest. As a preamble, I'm not super thrilled with the naming in this area. Others have complained. Naming is hard. I'm open to suggestions.
People get certifications, such as a driver's license. They get the certification in a particular Course, such as `PR driver's license` which specifies the term of the certification, etc. This gives them a skill, such as `Standard Driving`. This skill may be a requirement of a title, such as CERT member. 
The issue is that once the certification is expired, it should no longer count towards having that skill. It should still be shown, but it no longer counts.

Hope all that helped. Other questions ?

cesc1989:
Thanks @kgf I'll give it a deeper look and see whether I could work on this. I'll let you know.

kgf:
@cesc1989 Did you have any further questions ? Any updates ?

cesc1989:
Not right now. I'm working my way through the app and I'll let you know. Thanks for asking! @kgf

kgf:
Ok, thanks for your help.

PR #576, opened 2017-10-23
Expired cert should not grant skill

cesc1989:
This PR is a workaround for #15 

The idea is to only return skills for certifications which haven't expired.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -7,7 +7,7 @@ class Person < ActiveRecord::Base
 
   before_save :title_order
 
-  has_many :certs, -> { where("certs.status = 'Active'" ) }
+  has_many :certs, -> { where("certs.status = 'Active'") }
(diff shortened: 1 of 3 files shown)


cesc1989:
@kgf please take a look. I haven't added tests yet for this.

Screenshots of how this works:

1. Bad
bad

2. Good
good

Looking forward to your comments.

kgf:
This looks like a valid approach. Thanks for your help !
Kevin

cesc1989:
Cool thing. Then I'd need to make tests pass again and maybe add more.

cesc1989:
@kgf change it to an instance method so that `person_spec` don't break

Waiting for checks to pass.

kgf:
Thanks. I'm watching Travis !

kgf:
Thanks for your help,  @cesc1989

Issue #108, opened 2016-07-05
Create a Vagrantfile


kgf:
This will allow folks to use Vagrant to contribute to Lims

kgf:
This is not a top priority, but if it speaks to you....

cesc1989:
Hey, @kgf I was skimming through the issues list and found this. Is this still a thing? I use vagrant for my local environment and provision the machine with ansible. If you take a look to .gitignore you'd see I've ignored a bunch of files for this purpose.

I can unignore them and document how to use those files for this project. Let me know.

kgf:
I think many people use vagrant for local development. It would be nice to have that documented, in case others wanted to try it.

I want to make it as easy as possible for others to help !
Thanks
Kevin

cesc1989:
OK. I'll do it.

PR #579, opened 2017-10-25
Add Vagrant configuration

cesc1989:
This PR implements #108 

I took the job to organize a bit README file so that it looks better on web and added info about how to configure this project with Vagrant


kgf:
It will take me a little bit to go through that. I'm more familiar with Chef.
It seems you've added a roles directory, which stores the config for Ansible ? If that's true, could you add a README.md with a few lines to that effect ? You could even link to that from the main readme ?

Thanks for your help !

cesc1989:
You bet

kgf:
You know, perhaps you should move most of the Vagrant stuff there and link to it from the readme. Then people who are not interested in Vagrant only have to skip a single line, like `Follow this guide if you would like to set up Ready Responder using Vagrant`
What do you think of that idea, @cesc1989  ?

cesc1989:
LOL. I just ~uploaded~pushed a commit.

I was thinking about first when I was updating the README. Now you said it, makes more sense to me.

Count on it.

kgf:
Great, thanks.
I notice the play book calls for node.js... Where is that used ? (Pretty sure not in RR itself !)

cesc1989:
> Where is that used ?

Ruby on Rails needs a JavaScript runtime environment. One of the available and/or default is the ruby racer but it's way easier when NodeJS is configured in the machine. That's why.

See https://stackoverflow.com/q/7092107/1407371

kgf:
@cesc1989 , thanks for this PR, it's very helpful.

Kevin

PR #580, opened 2017-10-25
Issue/578/add messages to recipients

### LINKED ISSUE #578: Add messages to recipients
Currently we have notifications, and it's good. Each notification has many recipients. The notification also has channels, which define how to send the message. Currently, the only channel available is text message, but we're adding email. When a notification is activated, it walks through the recipients and looks up the content on how to reach them. We then connect to Twilio and send the message.

This issue is looking to capture the return values from the call to Twilio. There is already a Message class to contain this info. Each recipient will have many messages. Each message will belong to a recipient. 

This line is where you can first get a handle on the message.
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/services/message/send_notification_text_message.rb#L25

You can look here for further documentation. => https://www.twilio.com/docs/api/messaging/send-messages

Thanks for your help !

bgates:
closes #578, I think. It doesn't create a `channel_id` for the messages, but that doesn't seem to be used anywhere.


app/models/recipient.rb
```diff
@@ -7,6 +7,10 @@ def notify! twilio = Message::SendNotificationTextMessage.new
     target_number = person.phone
     target_number = "+1#{person.phone}" if /\A\d{10}/.match target_number
     msg = twilio.sms_send notification.subject, target_number
-
+    if msg.is_a? Twilio::REST::RequestError
+      messages.create(status: "error", body: msg.message)
+    else
+      messages.create(status: "sent", body: msg)
```
kgf: Should the body of the Message be the same in both cases ? ( `msg` vs `msg.message`).

kgf:
The channel id is referring to the database record for the number it was sent to.
https://github.com/ReadyResponder/ReadyResponder/blob/23a24667638f34d5be517cf68eff1513375899d9/app/models/phone.rb#L1
It would be nice to set it, as the UI will want it soon.
Thanks for your help !
Kevin

app/models/recipient.rb
```diff
@@ -7,6 +7,10 @@ def notify! twilio = Message::SendNotificationTextMessage.new
     target_number = person.phone
     target_number = "+1#{person.phone}" if /\A\d{10}/.match target_number
```
kgf, line 8: person.phone has the channel.id you're looking for.

bgates: Thanks!

app/models/recipient.rb
bgates: I was wrongly treating the success path as getting a string for `msg`. Looking at the API docs you linked earlier (https://www.twilio.com/docs/api/messaging/message) I see that a successful call to the API will return an object with a `body` attribute. For the error, though, the twilio gem converts the response into a subclass of `StandardError`, which has no `body` only a `message`.

kgf:
@bgates Thanks for your help !
Can I interest you in other changes in this area ?

bgates:
@kgf Sure, what else?

PR #584, opened 2017-10-26
Set default task status to active

hmasila:
Set default task status to active

There are two approaches involved:
1. Having a `before_save` snippet that sets the status to 'Active' if the status is not provided
2. Adding a default value in the db migration for tasks model

### DIFF
--- a/app/views/tasks/_form.html.erb
+++ b/app/views/tasks/_form.html.erb
@@ -4,7 +4,7 @@
   <div class="form-inputs">
     <%= f.input :title %>
     <%= f.input :priority, as: :select, collection: Task::PRIORITIES %>
-    <%= f.input :status, :as => :select, :collection => ['Active', 'Cancelled'], :selected => 'Active' %>
+    <%= f.input :status, :as => :select, :collection => ['Active', 'Cancelled'] %>
     <%= f.input :description %>
     <%= f.input :location %>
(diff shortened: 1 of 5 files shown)


app/views/tasks/_form.html.erb
```diff
@@ -4,7 +4,7 @@
   <div class="form-inputs">
     <%= f.input :title %>
     <%= f.input :priority, as: :select, collection: Task::PRIORITIES %>
-    <%= f.input :status, :as => :select, :collection => ['Active', 'Cancelled'] %>
+    <%= f.input :status, :as => :select, :collection => ['Active', 'Cancelled'], :selected => 'Active' %>
```
kgf: This was much closer to what I was expecting, but it would be better to have this in the controller.

hmasila: Let me try that then

kgf: Thanks !

db/migrate/20171026191711_change_column_default.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeColumnDefault < ActiveRecord::Migration
```
kgf: I would prefer to not put a default in the database.

kgf:
@hmasila Thanks for your help !

PR #585, opened 2017-10-26
[issue-450] added strong params to controllers

sskates1:
Also remove use of string in params since params are no longer hash with indifferent access.


kgf:
Thanks for your help, @sskates1

PR #590, opened 2017-10-30
removed todos for strong params

sskates1:


kgf:
@sskates1 Thanks for all your great work on this. 
Do you know if these models now have strong params in the controllers ?

kgf:
Thanks @sskates1 !

Issue #589, opened 2017-10-30
It would be nice to have a button to move from one event to the next


kgf:
The "Next" one would be defined by chronological order.
Previous would be good to make it consistent.

valterbarros:
Do I want to do this, ok?

valterbarros:
Where is the place for this button? on show from the event?

valterbarros:
Do what you think about this one?

`Event.where('start_time > ?', Event.first.start_time).order(:start_time).first`

kgf:
`Event.where('start_time > ?', Event.first.start_time).order(:start_time).first`
The problem I see is that it does not reference the one you're on. 
Right now we have an event for Halloween, and another for Friday. If I'm on the Halloween event, the `Next` link would take me to the Friday event; not the events in December.
]The button should be on the event#show action.
Does that make sense ?
Thanks for your help.
Kevin

valterbarros:
Yeah @kgf I know your question. But notice that I put a `first` to take the latest first event.

kgf:
In you PR, it references `self` which is what I expected.

valterbarros:
Hahah yeah, really

kgf:
It would be nice to use blue buttons similar to the blue + on the events page.
Thanks for your help!
Kevin

valterbarros:
I Created a new PR with the button. and functionality to jump next and the previous event.

PR #591, opened 2017-10-31
It would be nice to have a button to move from one event to the next #589

valterbarros:
- Create new method on model to find next and previous event
- Cover both by test

Signed-off-by: ValterBarros <valterinside@gmail.com>

### DIFF
--- a/spec/models/event_spec.rb
+++ b/spec/models/event_spec.rb
@@ -72,4 +72,30 @@ RSpec.describe Event do
     expect(Event.upcoming.include?(@event)).to be(true)
   end
 
+  describe '#next_event' do
+    let(:event_001) { create :event, start_time: "2017-11-02 03:36:00", end_time: "2017-11-04 03:36:00" }
+    let(:event_002) { create :event, start_time: "2017-11-03 03:36:00", end_time: "2017-11-06 03:36:00" }
+    let(:result) { event_001.next_event }
+    before do 
+      event_002
+    end
+
+    it 'should return event_002' do
+      expect(result).to eq event_002
+    end
+  end
+
+  describe '#previous_event' do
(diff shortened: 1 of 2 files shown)


valterbarros:
closes #589. Guys, I did create a part from backend and something is missing but how today is the last day from hacktorberfest I already up this pr to guarantee my t-shirt ;D. Trust me I will continue the work.

kgf:
@valterbarros No problem, we understand. I'll put a `On hold` label on it, keep us updated. I commented in the issue, #589

spec/models/event_spec.rb
```diff
@@ -72,4 +72,30 @@
...
+  describe '#next_event' do
+    let(:event_001) { create :event, start_time: "2017-11-02 03:36:00", end_time: "2017-11-04 03:36:00" }
+    let(:event_002) { create :event, start_time: "2017-11-03 03:36:00", end_time: "2017-11-06 03:36:00" }
+    let(:result) { event_001.next_event }
+    before do 
+      event_002
+    end
+
+    it 'should return event_002' do
+      expect(result).to eq event_002
```
kgf: These tests would be much more valuable if you added an `event_000` and an `event_003`. Think if someone changed the code to get the last event, not the next. These tests would still pass.
Also, if you put this in a context, your `let` statements could happen once, eliminating the Code Climate complaint.

valterbarros: Yeah, good.

kgf:
@valterbarros I have merged this because it does advance issue #589 , but it does not close it, because it does not include the UI portion. Perhaps you would be willing to take that on as well ?

valterbarros:
@kgf Yes I can do this

kgf:
It would be nice to use blue buttons similar to the blue + on the events page.
Thanks for your help!
Kevin

PR #592, opened 2017-10-31
auto assign person to requirement

### LINKED ISSUE #547: When someone is available, assign them if appropriate
Currently, an Event has tasks, a Task has requirements, a Requirement has assignments. An Assignment is the join between a Requirement and a Person, showing they are expected to fulfill that requirement. Currently, individual volunteers can use text messages to communicate with RR. This is great because then you don't need to manage logins. They can text in that they are available during an event.

What we would like is for requirements to have a new attribute, `auto_assign`, as a boolean. If true, the Person listed in the Availability will be checked to see if they meet the requirements of the Requirement. If they do, an additional check will be made to see if the Requirement is already full. If not full, an Assignment record should be made automatically.

The person should be notified of this, but this first effort does not need to include this. It is optional.

This first iteration can assume that only one Requirement will have auto-assign enabled. A second future issue will deal with the logic of how to fill multiple auto-assign requirements.

bgates:
closes #547, at least the first stage of it. 

Thanks for your explanation!

### DIFF
--- a/spec/services/msg/available_spec.rb
+++ b/spec/services/msg/available_spec.rb
@@ -42,6 +42,53 @@ RSpec.describe Msg::Available do
           expect(availability.end_time).to eq(@event.end_time)
         end  
       end
+
+      context 'when there is an auto-assignable requirement for the event' do
+        let(:requirement) { build :requirement }
+        before(:example) do
+          allow(requirement.assignments).to receive(:create)
+          allow(@event).to receive_message_chain('requirements.find_by') { requirement }
+        end
+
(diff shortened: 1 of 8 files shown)


spec/services/msg/available_spec.rb
```diff
@@ -42,6 +42,53 @@
           expect(availability.end_time).to eq(@event.end_time)
         end  
       end
+
+      context 'when there is an auto-assignable requirement for the event' do
```
kgf, line 46: I like how you laid out the new tests.

PR #594, opened 2017-11-01
prevent availability from starting in the past

### LINKED ISSUE #588: When a new availability comes in via text, it can't start in the past
This will make it easier for end users, so they won't have to use `Available custom`.
When one does come in, it should be set to `Time.zone.now`

bgates:
closes #588.


kgf:
Thanks very much for all your help @bgates . I hope you'll continue even after hacktoberfest !

PR #596, opened 2017-11-03
add-next-and-previous-functionality-add-button-next-to-view

valterbarros:
- Add next and previous button to view
- Create helper methods to take the next_event and previous_event from event model
- Fix false positive on tests add :desc option on model to previous_event method

Signed-off-by: ValterBarros <valterinside@gmail.com>

### DIFF
--- a/spec/models/event_spec.rb
+++ b/spec/models/event_spec.rb
@@ -77,10 +77,15 @@ RSpec.describe Event do
     let(:event_001) { create :event, start_time: 2.days.from_now, end_time: 4.days.from_now }
     let(:event_002) { create :event, start_time: 3.days.from_now, end_time: 7.days.from_now }
     let(:event_003) { create :event, start_time: 7.days.from_now, end_time: 9.days.from_now }
-    before { setup }
+
+    before do
+      event_000
+      event_001
+      event_002
+      event_003
+    end
 
     describe '#next_event' do
-      let(:setup) { event_002 }
       let(:result) { event_001.next_event }
 
       it 'should return event_002' do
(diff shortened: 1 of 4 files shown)


valterbarros:
closes #589

valterbarros:
Note that I created a helper method to make this more easy and simple, I thought to create a controller action but for this feature is not too necessary.

spec/models/event_spec.rb
```diff
@@ -77,10 +77,15 @@
     let(:event_001) { create :event, start_time: 2.days.from_now, end_time: 4.days.from_now }
     let(:event_002) { create :event, start_time: 3.days.from_now, end_time: 7.days.from_now }
     let(:event_003) { create :event, start_time: 7.days.from_now, end_time: 9.days.from_now }
-    before { setup }
+
+    before do
+      event_000
+      event_001
+      event_002
+      event_003
```
kgf: I think you have these lines because oitherwise the `let` statements won't happen ( lazy evaluation ) ? You can accomplish that by using `let!` in line 79, then the object will always be created. If that does what you want, we could drop this entire block.

valterbarros:
Updated

spec/models/event_spec.rb
valterbarros: ok

kgf:
Thanks for your help @valterbarros !

valterbarros:
Oh yeah, thank you.

valterbarros:
Can I into in your slack channel? I tried but seems that need an account to sign in.

kgf:
@valterbarros what email would you like to use? You can send it to kfallon@readyresponder.org

valterbarros:
Oks

PR #609, opened 2017-11-14
Initial OpenLayer map commit 

marshall-alex:
Work in progress. added new file maps.js to app/assets/javascripts. maps.js contains an example of JS for an OpenLayers map, it still needs to be configured. Added new div to tasks/show.html.erb that is dependent on existance of lat/long for task. Currently including script at line 34 of tasks/show.html.erb for openlayers but this is only temporary for testing


kgf:
Thanks for your help !

marshall-alex:
No problem Kevin! Excited to be working on this project. I managed to get a
little work done today and should be pushing a much more flushed out demo
of the map either tomorrow or Friday and we can see if it's what you had in
mind! Have a nice week.

best,
Marshall


Best regards,
      Marshall Alex
      805.886.1845

kgf:
This is quite exciting !

marshall-alex:
Just pushed a new commit that should be enough to see if we're on track. Let me know @kgf !

kgf:
I'm excited to take a look today !

Issue #610, opened 2017-11-14
edit availability erases person and changes person choice to 'Yes' 'No'


VanAuken:
When editing an availability, the Person field should still show original person.  Person field choices should be a list of active personnel. 
Currently, choosing edit and making any change erases the person's name and choices become blank, 'Yes' or 'No' and it fails validation.

yasirazgar:
Looking into it.

kgf:
@yasirazgar Thanks !
Adding on it what the OP said, the list of people should be all active, plus the original person, if editing. If someone becomes inactive,editing one of their availabilities would otherwise be impossible.

yasirazgar:
ok @kgf I will make sure that only active persons are displayed in the drop down, this issue occurs when there is some error while updating the availability and edit is rendered.

PR #614, opened 2017-11-19
#610 - Editing availability changes options in person drop down.

yasirazgar:
Issue - https://github.com/ReadyResponder/ReadyResponder/issues/610

Changes
1. Set all required variables before rendering edit, when there is an error while updating.


kgf:
@yasirazgar thanks for your help!

Issue #612, opened 2017-11-18
On inspection form, repairer is not limited to active people


kgf:
It is also not sorted by name

yasirazgar:
looking into it.

kgf:
Thanks for your help!

kgf:
Closed by #615

PR #615, opened 2017-11-20
#612 - limit repairer select in inspection form to active people

yasirazgar:
issue: https://github.com/ReadyResponder/ReadyResponder/issues/612

Description: Currently repairer select in inspection form shows all the persons including inactive people, it needs to show only active people.

Changes:
1. Added default sort logic, which sorts based on firstname, lastname and middleinitial to the person model.
2. Updated all places which uses old sorting logic for person to use new logic.
3. Updated repairer select in inspection form to show only active people.


kgf:
@yasirazgar Thanks a lot for this work. A nice PR !

Issue #392, opened 2017-04-25
Show multiple sets of status boxes for events


kgf:
This is referring to these boxes
![screen shot 2017-03-06 at 10 58 20 am](https://cloud.githubusercontent.com/assets/2431417/25407720/650c42ba-29d9-11e7-9fd6-eca63844b774.png)

In events, the status boxes are quite helpful for knowing how an event is staffed. Currently the event shows only one set, which can be inadequate for events with several departments responding. This issue is a request for one set **per department**. They should be stacked vertically, with the department name to the left. The department short name should replace the word `Personnel` in the screen shot above. They should drop below the event status.

kgf:
This may look slightly different now. There are now also blue boxes showing assignments, but it's still there same ticket.

firdausmaarof:
Hi. I'm interested to work on this one.

kgf:
@firdausmaarof that would be great, thanks. Any questions, feel free to post them here.

kgf:
@firdausmaarof Any questions or updates ?

kgf:
@firdausmaarof Are you still interested in working on this ?

kgf:
@firdausmaarof We haven't heard from you for a while. I'm going to set this back to `Help Wanted`. If you actually _are_ working on it, please post a comment and I'll reverse that action.

nesteves:
Hi @kgf, I would like to take a look at this.

kgf:
That would be great, thanks @nesteves

kgf:
@nesteves have you had any success on this issue?

nesteves:
@kgf had some trouble finding free time for this. I am picking it up again today and looking forward to finishing it no later than this weekend!

nesteves:
@kgf, just a small question, there are 5 status boxes:

* assigned
* available
* partially available
* unavailable
* no response

Are these mutually exclusive? I was working on that assumption writing feature specs for these, but when I create a person that has an assignment related to the event, making her a *assigned* person, it also affects the other 4 statuses, depending on whether I also define an availability, its status (*Available* vs *Unavailable*) and its length.

Thanks!
Nuno

kgf:
Assigned is a special case, it reflects when someone has something to do. The other four boxes are  mutually exclusive. You can be Available, partially available, unavailable, or unknown, but only one.
Thanks for your help.

kgf:
Fixed by #616, thanks to @nesteves

PR #616, opened 2017-11-25
Feature event status boxes

nesteves:
Implements the functionality requested in issue #392 

The event's page now shows one set of status boxes for each department:

rr_event_status

Summarizing the changes to the codebase:

* a bit of housekeeping on the files I worked with: deleted unused helper modules; tried to split the event feature specs by view or major piece of functionality; refactored the event and person model specs; added a spec to the assignment model
* refactored the methods that fetch an event's responses/people and upon which this feature depends so they could be chained as a normal relation. I chose not to unit test these as they call on other model scopes (_Availability_ and _Person_) that are already covered
* implemented the necessary changes to the view and helper


app/models/event.rb
```diff
@@ -51,51 +51,62 @@ def to_s
...
   end
 
   def responding_people
-    self.responses.map { |a| a.person }
+    Person.active.joins(:availabilities).
+      merge(Availability.containing(start_time..end_time))
   end
 
   def eligible_people
-    Person.active.where(department: self.departments).where("title_order <= ?", Person::TITLE_ORDER[min_title])
```
kgf, line 96: This change is incorrect, it should be `<=`.
1 is the most important title. 100 would be a less important title.
 Ideally we should have a test on this, as we've had this regression before.

kgf:
@nesteves Thanks for your help on this ! One small needed change, and a request for a test if you could...

Thanks !

app/models/event.rb
nesteves: Oh, so a person ranked higher than another should have a `TITLE_ORDER` that is *less than* the other person's.

Thanks for the heads up!

kgf:
@nesteves Thanks for the quick turn around. Sadly, the tests failed.

kgf:
Fixes #392 
Thanks for your help @nesteves !

PR #619, opened 2017-12-03
Changed two lines that used a newer syntax that was not supported in …

kgf:
…some production environments

Hoping to change these back at some point


kgf:
Sean, this was originally yours, hoping you have time to take a look. Thanks

kgf:
The codacy warning is due to the exact changes. It wants the newer syntax.

Issue #625, opened 2017-12-06
Unique_ids should include the following category choices


kgf:
This is to add Serial as a choice.
Correct values are
['VIN', 'Serial', 'Sub-component', 'Grant', 'Slug', 'Registration', 'Department', 'Other']

On this line
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/models/unique_id.rb#L7

kgf:
Fixed in #626

PR #626, opened 2017-12-06
Fix Issue #625: Unique_ids should include the following category choices

tungla:
Fix Issue #625


kgf:
@tungla Thanks for your quick work on this, it's appreciated !

Issue #632, opened 2017-12-12
Allow the qty of Upcoming events to be defined by a setting


kgf:
This should use the Settings model. It should do a find or create to look for the setting. If it creates, qty should be 10, Category should be Person, status active.

kgf:
Closed by #633

Issue #611, opened 2017-11-16
In Event, minimum title does not exclude lesser titles


VanAuken:
An Event with a minimum title of officer, will still show CERT members, may show recruits (uncertain).  If there is a simultaneous event for CERT, the second event will show CERT members as having not responded.

The Event should work that a minimum title officer will exclude all CERT members from showing up as available, unavailable, partially available and also not show as no response regardless if they are available at the same time.

This may be related to #606

kgf:
This is due to the comparison being the wrong direction. Higher ranked people have lower numbers.
The fix for this is easy, but I'd really appreciate a PR that included some tests !
Offending line is here. ->
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/models/event.rb#L94
Thanks for your help !
Kevin

yasirazgar:
@VanAuken @kgf, I cant understand the issue completely, can you guys please provide some additional details.
Thanks!

kgf:
The bug portion of this was fixed by #623. I will close this. 

This could use some tests though !Especially around the `unknown` box on the Event show

PR #633, opened 2017-12-17
632 - setting for upcoming events count

yasirazgar:
issue - https://github.com/ReadyResponder/ReadyResponder/issues/632

Summary:
Added new setting for defining upcoming events.


kgf:
Thanks for your help @yasirazgar ! I'll try to get time to look at this, but things are quite busy at the moment !

yasirazgar:
ok @kgf.
I need one more help, I am trying to work on this issue https://github.com/ReadyResponder/ReadyResponder/issues/611, 
but understand it properly, can you please take a look and add some more explanation once you are free. 
Thanks!

app/models/setting.rb
```diff
@@ -13,4 +13,22 @@ def self.get(key)
     value_object = Setting.where(key: key, status: "Active").first
     value_object.value if value_object
   end
+
+  def active?
```
kgf, line 17: I like this

app/models/setting.rb
```diff
@@ -13,4 +13,22 @@ def self.get(key)
...
+    self.status == 'Active'
+  end
+
+  def self.find_or_create_upcoming_events_setting(options = {})
```
kgf: I worry about this, because this direction will require one of these for every setting. To a large degree, I don't care if the setting is in the db. The biggest reason to always have them there is to let people know `Hey, there's an adjustment you can make here...`. We could add it to `seeds.rb`, which would make all this move there, keeping the model smaller. The seeds could create settings without values.

app/models/person.rb
```diff
@@ -229,7 +229,11 @@ def meets?(requirement)
   end
 
   def upcoming_events(count = 10)
-    self.department.events.limit(count)
```
kgf: I'd like to move this into a Setting class method. Something like 
`Setting.get_integer('Upcoming_event_count',10)`
This would look for the *active* setting. If it found one that had a non-null value, return that value. If it did not, return the default handed in (10).
The default should be optional. If there is no setting, and there is no default, return nil.
I hope to use Settings a lot in the app. I think this will be a more sustainable model. 
It may also make stubbing in tests easier.
Thoughts/concerns ?

app/models/setting.rb
yasirazgar: I was also doubtful about this, moving it into the seeds is correct option.

app/models/person.rb
yasirazgar: yeah totally agree, this should be inside the Setting model, sorry it was bad decision by me.

Will update the requested changes and let you know.

yasirazgar:
@kgf I have made the requested changes, please take a look when you have time.

app/models/setting.rb
```diff
@@ -13,4 +13,14 @@ def self.get(key)
     value_object = Setting.where(key: key, status: "Active").first
     value_object.value if value_object
   end
+
+  def active?
+    self.status == 'Active'
+  end
+
+  def self.get_value(key, fallback_value = nil)
```
kgf: This would be even better if we went to `get_integer`, as it would show intent even better. Then, as needed, we can add get_boolean , for example. These should always return the datatype or nil.

app/models/setting.rb
```diff
@@ -13,4 +13,14 @@ def self.get(key)
...
     value_object.value if value_object
   end
+
+  def active?
+    self.status == 'Active'
+  end
+
+  def self.get_value(key, fallback_value = nil)    
+    value = get(key)
+    (Integer value rescue value) || fallback_value
```
kgf: I'm not familiar with that form. Is that the same as `value.to_i` ?

app/models/setting.rb
yasirazgar: Yeah makes sense, do you also want me to add, get_boolean and get_string as well.

app/models/setting.rb
yasirazgar: Yes but `to_i` converts nil to 0, so I used this. If the value cannot be converted to the integer the raised exception will rescued and will return the actual value, if the value is nil, fallback value will be returned. 
Is this ok ?

yasirazgar: And now as we are sure that this method will only return integer, I think this can be changed to 
`(Integer value rescue nil) || fallback_value`. 
And if possible we also need to add validation to prevent non-integer value for this setting in the first place.

app/models/setting.rb
kgf: If you want, or we can wait til we need them.

app/models/setting.rb
kgf: I would appreciate a brief comment here, for others following along. Basically close to your comment above. I like to `resque nil` version
For validation, have a `set_integer` method ?

Thanks for your help !

yasirazgar: Sorry @kgf, what do you want me to comment about ?

yasirazgar:
@kgf I have made the changes please take a look when you have time.

app/models/setting.rb
```diff
@@ -13,4 +13,17 @@ def self.get(key)
...
+
+  def active?
+    self.status == 'Active'
+  end
+  
+  # this method should be used only for the settings that have integer values
+  # converts the given string to integer, if not possible returns nil.
+  def self.get_integer(key, fallback_value = nil)    
+    value = get(key)
+    (Integer(value) rescue nil) || fallback_value
```
kgf: Still struggling with this syntax some, and so is Code Climate. I tried it out in a console and it threw an error. What would you think of `value.is_a?(Integer)`

yasirazgar: yeah, that should also work, 
`Integer(value) rescue nil` is somewhat short. 
Code clarity is more important, I will change it.

yasirazgar: sorry @kgf, I think it wont work, because the type of the column, "value" is text, so 
`setting.value` will return the value in string format (example, '9'), so `value.is_a?(Integer)` will return false.
I cant think of any other way other than `Integer(value) rescue nil` as if the value can be converted it will convert and int value will be returned, if it cant, raised exception will be rescued and nil be returned.

We can use `value.to_i.to_s == value` but it wont work in some case like when value is '01'.
Or we can use regex not sure about that too.

Please let me know, if you have any better solution for this.

kgf: Phone a friend... BRB. Thanks for your diligence on this !

masukomi: presuming `get(key)` always returns a string... if it returns nil you'll have to add a check for that too .

```
def self.get_integer(key, fallback_value = nil)
  raw_value = get(key)
  value = raw_value.to_i 
  if value == 0 and !raw_value.match(/^\s*0+(\.0+)?\s*$/)
      value = fallback_value
  end
  value
end
```

yasirazgar: Wouldn’t it be better if we just use `Integer(value) rescue fallback_value`.
@kgf @masukomi what do you think ?, let me know your thoughts.

masukomi: Yeah, @yasirazgar I think you're right. Your line is atypical but i think it's relatively readable.

kgf: Yes, you've convinced me too. I don't care for @masukomi solution as much, because it looks like you could not have a setting of zero, as it would think that meant no setting.
Thanks to both of you for your work on this !

kgf:
Thanks for your help @yasirazgar ! I appreciate it.

PR #634, opened 2017-12-20
Alert Indicator: Staffing Level Endpoint

johnschoeman:
This creates an endpoint that responds with the current Event staffing level that will be used for the alert indicator device.

It is scoped under an api namespace to keep it seperate from the main application.

making a get request to ```/api/staffinglevel``` should return some json of the form: 
```
{staffing_level_number: 3, staffing_level_name: "Satisfied"}
```

jbuilder was added to the gemfile to help with the structuring of the json data.

Testing for the Event model was added in ```spec/features/alert_spec.rb```

Note that there is currently no authentication happening at the controller level.  I think this is okay for now since the endpoint doesn't allow for any mutations.  Authentication can be added later after the IoT end of this is more fully worked out.

I'm basing this off of the upgrade to rails 5 branch in anticipation to that migration.  It may make more sense to reject this pull request and I can re add the code to the main development branch after the upgrade branch has been merged into development.  This would keep the git history a little cleaner.  

Puma was added partly in anticipation for rails 5, but it will be required to use a server capable of multiple threads if action cable/websockets end up being used for live updating of the alert indicator.

I also ran across a few bugs in some of the controllers and the timecard model.  I added some fixes, but wasn't sure if that would introduce any conflicts with ongoing development, so I put them in separate commits in case you'd like to reject those commits specifically.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,6 +24,7 @@ gem 'twilio-ruby',             '~> 4.11.1'
 #gem "taps"
 #gem 'validates_timeliness', '~> 3.0'
 gem 'devise'
+gem 'jbuilder'
 
 source 'https://rails-assets.org' do
   gem 'rails-assets-datetimepicker'
--- a/app/assets/javascripts/api/alert.coffee
+++ b/app/assets/javascripts/api/alert.coffee
@@ -1,3 +1,0 @@
-# Place all the behaviors and hooks related to the matching controller here.
-# All this logic will automatically be available in application.js.
-# You can use CoffeeScript in this file: http://coffeescript.org/
--- a/app/assets/javascripts/api/alert.coffee
+++ b/app/assets/javascripts/api/alert.coffee
@@ -1,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -147,6 +147,41 @@ class Event < ApplicationRecord
     return true
   end
 
+  def self.staffing_level
+    begin
+      relevant_events = self.current_or_next_events
+      event_staffing = relevant_events.map { |event| event.staffing_level }
+      event_staffing.min
+    rescue => e
+      puts e
+      -1
+    end
+  end
+
+  def self.current_or_next_events
+    current_events = self.current_events
+    events = current_events.empty? ? [self.next_event] : current_events
+  end
+
+  def self.current_events
+    self.where("start_time < ?", Time.now)
+        .where("end_time > ?", Time.now)
+  end
+
+  def self.next_event
+    next_event = self.where("start_time > ?", Time.now)
+        .order("start_time ASC")
+        .first
+    raise '---Error: There are no events---' unless next_event
+    next_event
+  end
+
+  def staffing_level
+    tasks = self.tasks
+    raise '---Error: Event has no tasks---' if tasks.empty?
+    tasks.map { |task| task.staffing_value }.min
+  end
+
 private
   def calc_duration #This is also used in timecards; it should be extracted out
      if !(start_time.blank?) and !(end_time.blank?)
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -42,6 +42,14 @@ class Event < ApplicationRecord
 
   CATEGORY_CHOICES = ['Training', 'Patrol', 'Meeting', 'Admin', 'Event']
   STATUS_CHOICES = ['Scheduled', 'In-session', 'Completed', 'Cancelled', "Closed"]
+  STAFFING_LEVELS = {
+    500 => 'Error',
+    0 => 'Empty',
+    1 => 'Inadequate',
+    2 => 'Adequate',
+    3 => 'Satisfied',
+    4 => 'Full'
+  }
 
   def to_s
     title
@@ -150,12 +158,22 @@ class Event < ApplicationRecord
   def self.staffing_level
     begin
       relevant_events = self.current_or_next_events
-      event_staffing = relevant_events.map { |event| event.staffing_level }
-      event_staffing.min
+      event_staffings = relevant_events.map do |event| 
+        { staffing_number: event.staffing_level, 
+          staffing_percentage: event.staffing_percentage }
+      end
+      event_staffing = event_staffings.min_by { |staffing| staffing[:staffing_number] }
+      self.build_staffing_level(event_staffing[:staffing_number], event_staffing[:staffing_percentage])
     rescue => e
       puts e
-      -1
+      self.build_staffing_level(500, "NaN")
     end
+  end
+
+  def self.build_staffing_level(staffing_number, staffing_percentage)
+    { staffing_level_number: staffing_number,
+      staffing_level_name: STAFFING_LEVELS[staffing_number],
+      staffing_level_percentage: staffing_percentage }
   end
 
   def self.current_or_next_events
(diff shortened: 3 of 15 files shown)


johnschoeman:
Adding description of deliverables from slack conversation here for reference:

john [2:16 PM] 
To clarify what is meant by percentage of Full staffing:  this would be the number of persons that are currently available based on the Availability model over the total number of persons?


kevin_fallon [2:28 PM] 
No, nothing that simple !!


[2:28] 
I realize now that would be wrong anyway


[2:29] 
What we want is the worst condition from all of the tasks, and show that


[2:30] 
Tasks can be [Empty, Inadequate, Adequate, Satisfied, Full].


[2:30] 
I think we have methods for all those

[2:30] 
But we don't have an event method to roll them up

[2:33] 
Let's say I have a shelter event, with two tasks. Medical intake and parking. Each task has min, desired and max staffing. If assignments are zero, then empty. Above zero, but below min, inadequate. Above min, but below desired, adequate. Above desired, satisfied. At max, full.

[2:33] 
If medical intake were satisfied, but parking was empty, gauge should be red


[2:34] 
So, if you want to mock it, and stick to the IOT side, the method call should give you an integer, 0-4.

[2:35] 
The application does not fully account for timespans yet

john [3:24 PM] 
I see, so the endpoint would need to receive an event_id as params and would return an integer 0-4 based off the event's task staffing as you described?  and integer will dictate to the indicator what color to show? 

How would the indicator know what event it should be monitoring?

Mocking is a good idea, but I'm happy to do the rails side of this as well. (edited)


kevin_fallon [3:25 PM] 
The Event class should handle that


[3:27] 
So, on Event. have a method to return the status for the next event. Later, it would show the next event, or if an event is in progress, it would show the next operational period (edited)

[3:27] 
`Event.staffing_level` perhaps ?

[3:27] 
Naming can be hard

john [3:30 PM] 
so it is not possible to have multiple events happening at the same time?

kevin_fallon
[3:37 PM] 
It actually is, good point. Display one from each end for two? Or just show the worst task for all the events.

[3:37] 
Basically, if I look at this and it's green, all is well.


[3:39] 
I think that API call should return a level integer and a name, so I can use the same API call for Alexa


[3:40] 
{"staffing_level_index": 0, "staffing_level_name": "Empty"}


john [3:44 PM] 
okay.  to sum up.  we want an endpoint that when pinged returns some json data of the form
```{"staffing_level_index": 0, "staffing_level_name": "Empty"}```
 where the level is determined by the worst case event task for all currently happing events and if there are no currently happening events it should return the staffing level of the next event to happen?


kevin_fallon [4:02 PM] 
So far so good.


[4:02] 
Maybe a level of -1 is an error, show all blue leds

**Comment by kgf** [comment]
Further discussion landed on `{"staffing_level_index": 0, "staffing_level_name": "Empty", "staffing_level_percentage": 0}` as the JSON response

**Comment by kgf** [comment]
For V1, we have six possible responses from RR. The first is no response. The second would be error, represented by `{"staffing_level_index": 0, "staffing_level_name": "Error", "staffing_level_percentage": 0}`.
This probably means `Empty` needs to be index 1. Don't think a negative index would be good.

The next four could be Empty, Inadequate, Adequate, Satisfied and Full. We could translate these into percentages, if desired.

Empty and Inadequate should be red, Adequate should be yellow, and Satisfied and Full should be green.

If you look at a clock, Empty would show red from 7 to ~830. Inadequate would show red from 700 to 10:00. Adequate would show red from 7:00 to 10:00, and yellow from 10:00 to 2:00. Satisfied would show red from 7:00 to 10:00, yellow from 10:00 to 2:00, and green from 2:00 to ~ 3:30. Full would show all that, plus green from 2:00 to 5:00.

Error could be red from 5:00 to 7:00. 
No response could be white from 5 to 7, as that would be normal on start up (edited)

File: Gemfile [hunk 1]
```diff
@@ -24,6 +24,8 @@ gem 'twilio-ruby',             '~> 4.11.1'
 #gem "taps"
 #gem 'validates_timeliness', '~> 3.0'
 gem 'devise'
+gem 'puma'
```
kgf: What is the purpose of puma here ?

app/assets/javascripts/api/alert.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
```
kgf: Could you please remove these two empty files ? Thanks

app/controllers/assignments_controller.rb
```diff
@@ -21,6 +21,8 @@ def new
   end
 
   def edit
+    @assignment = Assignment.find(params[:id])
```
kgf: This is unneeded, as the load resource on line 3 handles it.

app/controllers/assignments_controller.rb
```diff
@@ -51,9 +53,9 @@ def set_assignment
     @assignment = Assignment.find(params[:id])
   end
 
-    def set_requirement
-      @requirement = Requirement.find(assignment_params[:requirement_id])
-    end
+  def set_requirement
+    @requirement = Requirement.find(params[:requirement_id])
```
kgf: Why do you not want to use the assignment strong params ?

http://edgeguides.rubyonrails.org/action_controller_overview.html#strong-parameters

app/controllers/certs_controller.rb
```diff
@@ -10,8 +10,10 @@ def show
     @last_editor = last_editor(@cert)
   end
 
- def new
-    @cert.person_id = (certs_params[:person_id]) if (certs_params[:person_id]).present?
+  def new
+    if params[:cert]
```
kgf: Is this needed ? I thought the `if (certs_params[:person_id]).present?` at the end would cover this ?

app/models/event.rb
```diff
@@ -147,6 +147,41 @@ def ready_to_schedule?(schedule_action)
...
+    end
+  end
+
+  def self.current_or_next_events
+    current_events = self.current_events
+    events = current_events.empty? ? [self.next_event] : current_events
+  end
+
+  def self.current_events
+    self.where("start_time < ?", Time.now)
```
kgf: We'll need to add a check on status here.

app/models/event.rb
```diff
@@ -147,6 +147,41 @@ def ready_to_schedule?(schedule_action)
...
+    next_event = self.where("start_time > ?", Time.now)
+        .order("start_time ASC")
+        .first
+    raise '---Error: There are no events---' unless next_event
+    next_event
+  end
+
+  def staffing_level
+    tasks = self.tasks
+    raise '---Error: Event has no tasks---' if tasks.empty?
```
kgf: Conceptually, not sure this is an error. It seems reasonable that a new event would have no tasks.

Gemfile
johnschoeman: Webrick doesn't support of action cable, but puma does.  If we'd like live updating of the indicator in the future, there will need to be a switch.  Until then we won't need it though, I will take it out.

app/models/event.rb
johnschoeman: Agreed.  Though it is not clear what the staffing_level of an event with no tasks should be.  is it full or empty?

app/models/event.rb
johnschoeman: okay.  possible statuses are: ['Scheduled', 'In-session', 'Completed', 'Cancelled', "Closed"].  I assume that cancelled and closed events are the only ones that should not be included?

app/controllers/assignments_controller.rb
johnschoeman: in some views there was no requirement key to build the strong_params from, leading to a server error.

app/controllers/assignments_controller.rb
johnschoeman: All of the changes in the controllers where responses to 500 level errors I was hitting in certain views.  I say we do not include any of these. whatever the appropriate fix is, it belongs in another PR.

johnschoeman:
creating a new pull request to address comments

PR #636, opened 2017-12-22
Staffing level endpoint

johnschoeman:
This creates an endpoint that responds with the current Event staffing level that will be used for the alert indicator device.

It is scoped under an api namespace to keep it seperate from the main application.

making a get request to ```/api/staffinglevel``` should return some json of the form: 
```
{staffing_level_number: 3, staffing_level_name: "Satisfied", staffing_level_percentage: 80}
```

jbuilder was added to the gemfile to help with the structuring of the json data.

Testing for the Event model was added in ```spec/features/alert_spec.rb```

Note that there is currently no authentication happening at the controller level.  I think this is okay for now since the endpoint doesn't allow for any mutations.  Authentication can be added later after the IoT end of this is more fully worked out.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -163,13 +163,13 @@ class Event < ApplicationRecord
     end
   end
 
-  # def next_event
-  #   Event.where('start_time > ?', self.start_time).order(:start_time).first
-  # end
+  def next_event
+    Event.where('start_time > ?', self.start_time).order(:start_time).first
+  end
 
-  # def previous_event
-  #   Event.where('start_time < ?', self.start_time).order(start_time: :desc).first
-  # end
+  def previous_event
+    Event.where('start_time < ?', self.start_time).order(start_time: :desc).first
+  end
 
   def staffing_level
     {
@@ -193,9 +193,9 @@ class Event < ApplicationRecord
   end
 
   def calc_duration #This is also used in timecards; it should be extracted out
-     if !(start_time.blank?) and !(end_time.blank?)
+    if !(start_time.blank?) and !(end_time.blank?)
       self.duration = ((end_time - start_time) / 1.hour).round(2) || 0
-     end
+    end
   end
 
   def trim_id_code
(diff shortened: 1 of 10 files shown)


kgf:
@johnschoeman We have merged the Rails 5 code into development. Could you please re-point this PR towards development ?
Thanks for your help.
Kevin

johnschoeman:
@kgf rebased onto development.  It would probably make sense to revisit the api and implementation as well.

johnschoeman:
Especially since it's not passing any ci checks

kgf:
So, should we close this PR ? Will you (hopefully) open a new one ?

kgf:
This PR is of more interest. Will you be able to continue with it ?

johnschoeman:
@kgf , I went over the code again here and cleaned up some stuff, added a few tests.

I think it is good to rebase and merge. We may want to add some better factories and add an ssh token or something to handle basic auth directly after merging this.  

Sorry it took me a while to get back around to this by the way.  I should have a bit more availability moving forward.

app/models/event.rb
```diff
@@ -163,11 +171,31 @@ def previous_event
...
+  end
+
+  # TODO: update this once staffing % meaning is clarified.
+  def staffing_percentage
+    staffing_number * 20 + 20
+  end
+
   def calc_duration #This is also used in timecards; it should be extracted out
-     if !(start_time.blank?) and !(end_time.blank?)
+    if !(start_time.blank?) and !(end_time.blank?)
```
gasparyn, line 196: The `and` here might give us unintended consequences because precedence. 
Additionally, it's better to use `if start_time.present? && end_time.present?`

app/models/staffing_level.rb
```diff
@@ -0,0 +1,48 @@
...
+  def build_staffing_level_error(e)
+    {
+      staffing_level_number: 500,
+      staffing_level_name: "Error: #{e}",
+      staffing_level_percentage: "NaN"
+    }
+  end
+
+  def current_or_next_events
+    if !current_events.empty?
```
gasparyn, line 25: I'd do something similar here:
`if current_events.present?`

app/models/staffing_level.rb
```diff
@@ -0,0 +1,48 @@
...
+      current_events
+    elsif !next_event.empty?
```
gasparyn, line 27: similarly: `elsif next_event.present?`

PR #637, opened 2018-01-03
Removed birthdate from QR code

MattW216:
Hi, @kgf I believe I've removed the birthdate from the QR. Please review when ready.


kgf:
Welcome @MattW216 ! Thanks for your help.

Issue #613, opened 2017-11-18
On item index view, show last inspection


kgf:
It would be good to know what needs an inspection.

saltmade:
@kgf Interested in taking this one on! Can I get access to the slack for relevant questions?

kgf:
Sure, what email for invite ? 
I certainly want to meet you where is easiest for you, but I also like to have info here so others can follow along.

saltmade:
@kgf jeff@jgheaton.com, but I'll ask questions here then!

saltmade:
@kgf Is 'last inspection' going to be a required field?

kgf:
Last inspection will actually not be a field at all. Inspection belongs to an item. An item has many inspections. So you would want to look up the inspections for the item, and give the most recent date from those inspections. Watch for N+1 queries. Also, inspections could be cancelled, so don't count those.
Thanks for your help!

saltmade:
@kgf Does it make sense to include the last inspection on `/items` since it's not in the navigation? Should it be in `/item-types/:id`?

kgf:
You are correct ! Good catch !

jefheaton:
Addressed in https://github.com/ReadyResponder/ReadyResponder/pull/638

kgf:
Closed by #638 
Thanks for your help.

PR #638, opened 2018-01-03
Added 'last inspection' to item_type index

saltmade:
Issue: https://github.com/ReadyResponder/ReadyResponder/issues/613
Adds the last inspection date of a given item to the item_type show page, where individual items are displayed. 

Last_inspection_date query added to model.

Testing added in `features/inspection_display_spec.rb`

### DIFF
--- a/app/views/items/_items_table.html.erb
+++ b/app/views/items/_items_table.html.erb
@@ -7,6 +7,7 @@
       <th>Status</th>
       <th>Condition</th>
       <th>Location</th>
+      <th>Last Inspection</th>
       <th></th>
     </tr>
   </thead>
@@ -19,6 +20,7 @@
         <td><%= item_status_label(item) %></td>
         <td><%= item_condition_label(item) %></td>
         <td><%= item.location.name if item.location %></td>
+        <td><%= item.last_inspection_date %></td>
         <td><%= table_button_link 'Edit', edit_item_path(item) unless cannot? :edit, item %></td>
       </tr>
     <% end %>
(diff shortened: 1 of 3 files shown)


spec/features/inspection_display_spec.rb
```diff
@@ -37,6 +37,15 @@
     end
   end
 
+  context "on item index" do
+    let(:an_item) { create(:item, item_type_id: 1) }
+    let!(:an_inspection) { create(:inspection, item: an_item) }
```
kgf: Since you only create one inspection, there is no assurance that you're testing getting the most recent inspection.

app/views/items/_items_table.html.erb
```diff
@@ -19,6 +20,7 @@
         <td><%= item_status_label(item) %></td>
         <td><%= item_condition_label(item) %></td>
         <td><%= item.location.name if item.location %></td>
+        <td><%= item.inspections.first.inspection_date %></td>
```
kgf: I'm not requiring it, but what do you think of having an `@item.last_inspection_date` to incapsulate this in the model ?

saltmade:
@kgf Changed to a method on the model! I think this obviates the need for the `includes` call also.

spec/features/inspection_display_spec.rb
```diff
@@ -37,6 +37,17 @@
     end
   end
 
+  context "on item index" do
+    let(:an_item) { create(:item, item_type_id: 1) }
+    let!(:an_inspection) { create(:inspection, item: an_item, inspection_date: 1.day.ago) }
+    let!(:an_older_inspection) { create(:inspection, item: an_item, inspection_date: 2.days.ago) }
```
kgf, line 43: I like how you have an older one that should not be found.

kgf:
Fixes #613 
Thanks for your help!

PR #644, opened 2018-01-22
Fix the bug that show the cancelled events on the index page

### LINKED ISSUE #603: Events calendar should not show cancelled events
When you click on Events in the top menu, you go to the events page. This page has a calendar at the top, and a table of events below that. If you have a cancelled event, it shows on the calendar, but does not show on the table. We do not want it to show on the calendar either.

ozmar-11:
It closes #603 

Modify the recent scope in the in the event model to avoid bring the Cancelled events.


kgf:
I've been pondering this PR for a while. It's simple enough to understand, and completely closes the issue. Thanks @ozmar-11 !
But I am not completely comfortable with changing the `recent` scope to now consider status. I think there should be a second scope regarding status, and then chain them together. 
However, I can not think of a good name for this scope. `not_cancelled` is descriptive, but not exactly smooth...
Can you come up with a better name for the scope ? Naming is hard.

ozmar-11:
Hi @kgf I think you are right the changing of the scope issue, and the name `not_cancelled` sounds good, but also we can use something like `except_cancelled` or `except('Cancelled')`, which one do you prefer?

kgf:
I would prefer `except_cancelled`

Issue #601, opened 2017-11-09
On requirement#show, display auto-assign attribute


kgf:
It should be placed down near `optional` attribute.

MattW216:
Hi, @kgf I'd like to take this one...

kgf:
That would be great, thanks ! Let me know if you have any questions

kgf:
Fixed by #663

PR #663, opened 2018-03-03
601 display auto assign attribute

MattW216:
Hi, @kgf Is this what you were looking for? Or would you like it in a different location? Let me know.

### DIFF
--- a/config/database.example.yml
+++ b/config/database.example.yml
@@ -3,8 +3,8 @@ development:
   encoding: unicode
   database: lims_development
   pool: 5
-  username: matt
-  password: B00thbay
+  username: <%= ENV['matt'] %>
+  password: <%= ENV['B00thbay'] %>
   min_messages: WARNING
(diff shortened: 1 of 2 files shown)


config/database.example.yml
```diff
@@ -3,21 +3,22 @@ development:
   encoding: unicode
   database: lims_development
   pool: 5
-  username:
-  password:
+  username: <%= ENV['matt'] %>
```
kgf: I will not want any of the changes to this file...
And you may want to change your password!

kgf:
Thanks for your help, comments made.

MattW216:
I'm sure it's just me, but I've just looked through each of the commits above, and at the conversation on the Issue, but I don't see any comments...

kgf:
Sorry, GitHub had it as a pending review, instead of a simple comment. You should be able to see it now.

MattW216:
Should be all set. Sorry about that - misunderstood the set-up instructions.

kgf:
Thanks for your help, @MattW216 
Fixes #601

Issue #199, opened 2016-10-15
Enable messages to be sent via email


kgf:
This may need to wait for messages to be created.
https://github.com/ReadyResponder/ReadyResponder/wiki/Notifications

This would need to add a unique hash to the subject line, so that Ready Responder can link it into the correct conversation.

Spandana-Sai:
@kgf I think this in interlinked with #200. Can I create new models and start working on it?

Spandana-Sai:
Do you have any mock up for the email template? @kgf

kgf:
Not so much. It doesn't need to be beautiful on the first pass. It should include the event id_code as a way to relate it to what they are available for. The event title and start and end times should be in there.
If you get the frame work of sending a message, future PR s can refine it.
Thanks for your interest!
Kevin

kgf:
Current production systems are on AWS, which grants 60,000 msgs/month free.  Therefore, using the aws-ses gem on the first pass would be fine. 
Eventually I would like to support other email providers, such as sendgrid or mailgun. Mailgun would be a front runner there, as it also has a free tier, which Sendgrid does not.

kgf:
@ivanfrias would this be of interest ?

ivanfrias:
Sure!

Em 12/11/2017 2:02 da tarde, "Kevin Fallon" <notifications@github.com>
escreveu:

kgf:
@ivanfrias Haven't heard from you in a while. @pcai has expressed interest in this issue, so I'm going to assign it to him. 
If you're still willing to help with the project that would be great, just let me know.
Thanks
Kevin

kgf:
I spoke with Peter recently about this issue. We have identified at least two email providers who offer a generous free tier for email. One is Mailgun, which offers 40K free emails per month. Amazon SES offers 1M free per month. I would prefer not to lock into a specific vendor, but I expect many people will want to deploy to AWS.
I believe you can conditionally load a gem based on a environment variable. We may also be able to forgo the gems, and simply use ActionMailer with basic SMTP.

kgf:
Thanks to @pcai , a good bit of this work has been done. However, he does not have time to work on this further at this time. 
As a first pass, I would like to use AWS SES for mail delivery. Using the gem is fine. A second iteration/issue could add a second option.

pcai:
i would be happy to review a pr though

PR #665, opened 2018-03-08
hook up actionmailer to some environment variables to send email via smtp

pcai:
This PR is part of https://github.com/ReadyResponder/ReadyResponder/issues/199

Adds dotenv gem 
Adds appropriate gitignore changes
Adds SMTP email support, configurable via environment variables

Here's a snippet of my `.env` in my development environment, using a sandbox mailgun account:
```
MAIL_METHOD=smtp
MAIL_SMTP_ADDR=smtp.mailgun.org
MAIL_SMTP_PORT=587
MAIL_SMTP_DOMAIN=sandbox510cbc9b5a074f459e51eb176e7d54d7.mailgun.org
MAIL_USERNAME=dev@sandbox510cbc9b5a074f459e51eb176e7d54d7.mailgun.org
MAIL_PASSWORD={not displayed here}
```

I propose this PR as it is a significant first step towards any email notification system, and I recommend this because:
* ActionMailer can do most of the heavy lifting. In fact, devise hooks into ActionMailer so I was able to send myself a password reset email without problems, in development.
* This setup is completely provider-agnostic - theoretically we could switch between mailgun, sendgrid, SES, or any number of providers by changing config variables.

A couple drawbacks to not using a native gem right away:
* Performance, if the difference between getting an email to inbox in 5s vs 10s is significant. I don't get the sense that it is. Using the SMTP protocol is generally slower performance-wise than the equivalent REST api. 
* `content-transfer-encoding` (etc) and other warts of the SMTP protocol sometimes leak through the abstraction.
* To process inbound emails or events from our provider, we need provider-specific logic anyway, since these tend to be in the form of webhooks (new inbound email, email clicked, email opened, etc).


config/environments/development.rb
```diff
@@ -39,12 +39,13 @@
...
+  # Use :letter_opener (open in browser), or :smtp (to test using real SMTP)
+  # Must be a symbol or it wont override action_mailer defaults.
+  config.action_mailer.delivery_method = ENV.fetch("MAIL_METHOD", "letter_opener").to_sym
   config.action_mailer.smtp_settings = {
-    address: "smtp.gmail.com",
-    port: 587,
-    domain: "billericaema.org",
+    address: ENV["MAIL_SMTP_ADDR"],
+    port: ENV["MAIL_SMTP_PORT"],
+    domain: ENV["MAIL_SMTP_DOMAIN"],
```
houhoulis, line 48: 👍

kgf:
@pcai, thanks for your help.
Is it your expectation that this PR would not have any effect on production sites, as it's part of the solution?
Kevin

kgf:
@pcai This PR looks good, but I was hoping you could answer the question above before merge.

pcai:
Sorry I missed your prior message. My expectation is that it will have no effect on production sites, given the following assumption:

* production sites currently do not send emails of any kind.

I am making this assumption because there are no settings in `production.rb` that I can see that would allow production to send emails

kgf:
Thanks @pcai

PR #670, opened 2018-03-11
Don't specify insecure 'git:' gem src in Gemfile.

houhoulis:
Fixes security issue that `bundle install` is warning about.


kgf:
Thanks, @houhoulis !

PR #672, opened 2018-03-11
Upgrade ruby patch version. Add 2.5 to .travis.yml

houhoulis:
Ruby 2.3.x patches include security fixes.  2.3.6 is the latest.

Have Travis run our tests in Ruby 2.5 as well, to help prepare for future upgrades.


houhoulis:
app doesn't boot on ruby 2.5 on travis. investigating

houhoulis:
We can't run RR on Ruby 2.5 until we upgrade '`devise`' to >= 4.4.0.

Upgrading '`devise`' that far will require updating some of our devise-related code such as `ApplicationController#configure_permitted_parameters`.

There are several gem upgrades in this PR that will also be necessary or desirable to run on Ruby 2.5, but I'd suggest throwing this PR out and starting over.  Consider removing Gemfile.lock and starting fresh with many updated gems.  Even with rails still locked to 4.2, 16 gems (not in this PR) could each be updated one or more _major_ versions.

houhoulis:
Ruby 2.3.6 is a good idea, but we're not ready for Ruby 2..5

PR #674, opened 2018-03-13
Availability fixes

masukomi:
Event#responses now sorted by status, start_time

changed what's passed to availabilities_table

it was passing all of the responses (partial and full) and then
passing in the partial responses (again) and then adding them together
which resulted in duplicate rows for anyone with a partial availability.

added Event#full_responses... and then didn't use it.

added tests for Event#partial_availability

partial_availability now gets folks who can only work middle


app/views/events/show.html.erb
```diff
@@ -74,11 +74,13 @@
...
+                                Assignment.active.for_time_span(@event.start_time..@event.end_time)).to_a
+                              %>
+
         <%= render(:partial => "availabilities/availabilities_table",
-                   :locals => { :availabilities => ( @event.responses +
-                                Assignment.active.for_time_span(@event.start_time..@event.end_time)),
-                                :partial_availabilities => ( @event.partial_responses)} ) \
-                                 || "No Availabilities yet !" %>
+                   :locals => { :availabilities =>
+                                responses_and_assignments.presence || "No Availabilities yet!"})%>
```
houhoulis: I think here, if `responses_and_assignments` is empty, you are passing a string but you need to pass an empty array.

houhoulis: Also: We're doing a lot of variable definition & logic in this view, rather than in the controller(s).  Logic in controllers is easier to untangle (IMO), and definitely easier to test. Not a blocker, but a personal preference.  (Even in this kind of situation where the availabilities_table partial is going to be rendered by three different views.)

kgf:
@masukomi Does this include/replace the changes we made during the event for seeing partial availability ?

kgf:
Sorry @masukomi this seems to break a lot of tests

masukomi:
will rebase to make mergeable

kgf:
Thanks

PR #677, opened 2018-03-15
You can still use mailcatcher - just not via Gemfile

houhoulis:
To use mailcatcher, run it as a stand-alone tool.
See https://mailcatcher.me :

> Please don't put mailcatcher into your Gemfile. It will conflict with
> your applications gems at some point.

While working on upgrading our gems, I am indeed seeing gem
dependency conflicts with mailcatcher, so this commit removes it
from our Gemfile.


kgf:
This has the added advantage of removing a lot of other dependencies. Thanks @houhoulis

Issue #487, opened 2017-10-02
An Item should know its inspection questions


kgf:
Currently an item can have an inspection, but it's kind of lame. All that's asked is the date and the current condition. What would be good would be to have a dynamic list of questions.
I would like the list of questions to be defined for each item_category, and also each item_type. In the future, an item may have more than one item_type in a tree structure. (An item_type may have other item_types as children.) The questions should be collected up and presented to the user on the inspection form.
For this iteration, all questions can be yes/no.

If this is too large for one PR, it could be broken into several steps. The first would be to have `@item.inspection_questions` return a collection of questions.

kgf:
@tegon, would this be of interest ?

tegon:
@kgf This one seems too big, I don't think I'll have time to finish it. But thanks anyway.

kgf:
@tegon thanks for your interest. Let me know if you'd like me to come up with others. It's an active project.

ChanChar:
@kgf I've added the `inspection_questions` association but going to need some clarification the rest of the task.

kgf:
I think that may be the wrong plan. The inspection questions will need to be on ItemCategory and ItemType. An Item will look up it's hierarchy, and gather all the questions to put on the Inspection. An Inspection instance will have something like `inspection_answers`, which should include the text of the original question, and the response. The first iteration can have simple pass/fail yes/no questions, like "Are the tires properly inflated ?" and "Do all the exterior lights work ?"
The answer should have more than a pointer to the question, in case someone edits the text of the question to have a different meaning.
Hope this helps.
Thanks
Kevin

kgf:
@ChanChar Any updates ? Anything I can help with ?

ChanChar:
@kgf Where are these questions coming from? 

> The answer should have more than a pointer to the question, in case someone edits the text of the question to have a different meaning.

Shouldn't we simply remove the associated answer and have the user reset the answer if this is a concern?

All in all, I may not have time to do this and will leave it open for someone else.

kgf:
OK, no problem. I'll just reset the labels. Thanks for your help in the past !

nsantiago2719:
Hi @kgf , I think I can take this. But for just clarifications, what you want is, the user should be able to add a dynamic question on each question right?

kgf:
@nsantiago2719 Thanks for your interest. 
I don't follow quite what you mean. Be sure to read my comment from ten days ago. 

[...]

PR #682, opened 2018-03-22
[WIP] Jheaton/i487/an item should know its inspection questions

saltmade:
This is a 🚧WIP commit for introducing inspection questions through tagging.

### DIFF
--- a/app/controllers/inspection_tags_controller.rb
+++ b/app/controllers/inspection_tags_controller.rb
@@ -1,0 +1,25 @@
+class InspectionTagsController < ApplicationController
+  def index
+    @inspection_tags = InspectionTag.all()
+  end
+
+  def new
+    @inspection_tag = InspectionTag.new()
+  end
+
+  def create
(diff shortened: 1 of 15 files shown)


saltmade:
@kgf Still working on this one, but I've got some more time today and tomorrow so I'll see if I can't get into the real questions, which look like:
1. How does the UI handle associating questions and tags?
2. How do we roll up those questions based on tags?

I started with this many to many, polymorphic tag relationship but I'm not convinced I won't have to go back and refactor it depending on how gnarly the queries get. That said I realized the best way to figure that out was by doing so 🤷‍♂️

kgf:
Thanks !
I was asking some associates, they suggested `acts as taggable`. Don't know if you took a look at something like that...

kgf:
This is regarding issue #487.
@jefheaton Are you still looking at this ?

saltmade:
@kgf I am not unfortunately

Issue #679, opened 2018-03-20
Add tests to event#show availabilities table


kgf:
We have had recent examples of this table not showing people. There was a change that attempted to fix this, and then we got duplicates.
This issue is only looking for the tests.

kgf:
This is the line that renders the table of interest
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/views/events/show.html.erb#L77

kgf:
When setting up the test cases, you'll want multiple availabilities 

1. One that starts before the event and ends during the event (Counts as partially available)
2. One that starts during the event, and ends after the event. (Counts as partially available)
3. One that starts and ends during the event (Counts as partially available)
4. One that starts before the event and ends after the event (Counts as fully available)
5. One that starts and ends before the event (Should not appear in the table)
6. One that starts and ends after the event. (Should not appear in the table)
7. One that starts and ends at the same times as the event (Counts as fully available)

Ideally there should be one or more availability records showing available for each of those, and one or more availability record set to unavailable for each of those time periods.

johnschoeman:
ping: @houhoulis & @kgf 

<img width="1272" alt="screen shot 2018-03-20 at 1 34 17 pm" src="https://user-images.githubusercontent.com/16049495/37801397-ddc20adc-2dfc-11e8-8098-a7dab65c0ede.png">


done with the first pass at these tests.  see [PR](https://github.com/ReadyResponder/ReadyResponder/pull/683)

I wasn't sure of the intended behavior, since the feature is currently not working.

Questions: 

- are partial availabilities supposed to be showing on the availability table?  If so, how are they indicated as partially available?

- are availabilities that have the status as unavailable supposed to show up on the table if the start/end times are such that the availability is considered full or partial?

- currently there is test coverage on the model level and controller under the context of 'responses'.  What is the distinction between a response and an availability?  do these tests need to be looked at again?

kgf:
John, thanks for your help !
Yes, partial availabilities should show up on the availabilities table. they are designated by a yellow background. Fully available is green, and unavailable is red. 
In addition to availabilities, assignments also show in this table as blue. This is to help those scheduling to see who already has an assignment. There are almost certainly better ways to do this, but that's what we have for now.
No matter the time span, unavailable shows as red.
Responses are availabilities. Wether you said you were available or unavailable, you have responded.
I hope this helps, and I'll be about today if you have more questions.
Thanks again,
Kevin

[...]

PR #683, opened 2018-03-22
Js add specs for availablities on event show page

johnschoeman:
[issue 679](https://github.com/ReadyResponder/ReadyResponder/issues/679)

This adds a feature spec and request specs for viewing the correct availabilities in the Availabilities table on events#show.

The feature spec runs on my machine in about 12 seconds
The request specs run on my machine in about 9 seconds

given there are about 8 cases that were wanted to be tested, I didn't want to use feature specs for all of them because that would have added about 96seconds to the test suite.

currently the feature spec is failing. and the request specs that test partial availability are also failing.

I don't have a lot of confidence in these tests since the app is currently not showing availabilities correctly, I'm not sure what the intended behavior is exactly or if the tests will pass properly once the main feature is fixed. 

I would suggest not merging this in yet until the tests have had some verification that the are testing the intended behavior.  I am creating this PR to start that discussion.

### DIFF
--- a/spec/requests/events_spec.rb
+++ b/spec/requests/events_spec.rb
@@ -1,0 +1,81 @@
+require 'rails_helper'
+require 'support/request_helper'
+
+RSpec.describe "GET events/:id", type: :request do
+  describe "Availablities table shows correct availablities" do
+    before(:each) do
+      login_as_editor
+
+      @event_start_time = 2.hours.from_now
+      @event_end_time = 12.hours.from_now
+      @event = create(:event, start_time: @event_start_time, end_time: @event_end_time)
+
+      @two_hours_before_event_start = @event_start_time - 2.hours
+      @two_hours_after_event_start = @event_start_time + 2.hours
+      @two_hours_before_event_end = @event_end_time - 2.hours
+      @two_hours_after_event_end = @event_end_time + 2.hours
+    end
+
+    context "when the availablites are partially available" do
+      it "returns availabilities that start before the event and end during the event" do
+        availability = create(:availability, start_time: @two_hours_before_event_start, end_time: @two_hours_after_event_start)
+
+        get "/events/#{@event.id}"
+
+        expect(response.body).to include(availability.person.name)
+      end
+
+      it "returns availablities that start during the event and end during the event" do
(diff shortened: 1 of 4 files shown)


spec/requests/events_spec.rb
```diff
@@ -0,0 +1,81 @@
+require 'rails_helper'
+require 'support/request_helper'
+
+RSpec.describe "GET events/:id", type: :request do
+  describe "Availablities table shows correct availablities" do
+    before(:each) do
+      login_as_editor
+
+      @event_start_time = 2.hours.from_now
```
johnschoeman, line 9: I'm not a fan of this before each block.  It would be better to put this logic in the tests themselves, but there would need to be work done on the availabilities factories to make the code clean.  I was running out of time that I could put on this though.  I can come back to this at a later date.

kgf: Is your concern around line 9 specifically ?

johnschoeman: no, the whole before block is a little brittle.  I think the next step for these tests are to make some more expressive factories, but I wanted to be clear on the intended behavior before moving forward on that.

kgf: Thanks for your help.
If you get to the factories, I have a preference that they not create associated objects. I have seen that become a cascade of things being created that are not needed for the test at hand. Associated objects can be created in the spec itself.

johnschoeman: Agreed, that is a better practice.

johnschoeman:
@kgf Also, I've gotten a bit busy I'm not sure if I'll be able to come back to this before the weekend.  How time critical is getting this issue resolved?

kgf:
I believe we have a solution for the bug already, but we've had multiple regressions on that table. Thus the desire for tests coverage. If you can't get to it until the weekend, that's fine, thanks for the help !

kgf:
@johnschoeman Are you still working on this ? It may be less necessary after PR #705 
Please let me know,
Thanks for your help.
Kevin

kgf:
I'm going to close this for now. @johnschoeman if you would like it re-opened, please let me know !

Issue #687, opened 2018-04-03
new_item_type page errors when accessed through admin panel


saltmade:
**Description** :octocat:
When trying to add a new item type through the admin panel, the page errors out because there is no 

**Reproduction guide** 🐞
Start ReadyResponder
go to admin panel at the top
Click "Item Types"
Click New

**Observed behaviour**: 👀 💔
<img width="893" alt="screen shot 2018-04-03 at 11 15 16" src="https://user-images.githubusercontent.com/3383824/38258387-541eb0e0-3730-11e8-973d-fcb3f7e8ea19.png">

**Expected behaviour**: ❤️ 😄
See new item type page with option to choose a category

**System Info** 💻
OS: OSX
ReadyResponder: 2.8.0

yasirazgar:
Fixing it.

PR #689, opened 2018-04-10
687 - new item type page error

yasirazgar:
Issue: https://github.com/ReadyResponder/ReadyResponder/issues/687

## Change summary
Reason:
For item type there are two routes
1. One with item_category new, from item_categories, http://localhost:3000/item_categories/1/item_types/new
2. Without item_category, from item_types/new

2 Was causing errors, as action `new` always searches for item_category.

Fix: 
Fixed it by searching for item_category only if item_category params are present.

Also added missing tests for item_types_controller, and added redirect for delete action.

### DIFF
--- a/app/controllers/item_types_controller.rb
+++ b/app/controllers/item_types_controller.rb
@@ -12,7 +12,7 @@ class ItemTypesController < ApplicationController
 
   def new
     @item_type = ItemType.new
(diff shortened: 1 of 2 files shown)


kgf:
@yasirazgar Thanks for your continued help !

Issue #667, opened 2018-03-08
Event fails to find eligible people if min_title not defined


masukomi:
When you just need anyone you can get your hands on there's no realistic `min_title`  so `eligible_people` should return everyone who is active.... period.

this also affects the "No Responses" section of the Events "Availabilities" tab. 

2 possible solutions

Recommended: 
```
# Person
  scope :titled_equal_or_higher_than, lambda { |title|
    if ! title.blank?
      where('title_order <= ?',
            Person::TITLE_ORDER.fetch(title, DEFAULT_TITLE_ORDER))
    else
      where(nil) # there may be a better way to make a no-op scope in rails 4+
    end
  }
```

Alternative:
```
# Event
  def eligible_people
    people = Person.active.of_dept(departments)
    if ! min_title.blank?
      people = people.titled_equal_or_higher_than(min_title)
    end
    people
  end
```

kgf:
I would prefer to change these to not have negative logic. 
So, for the alt version
if min_title.blank?
  people = Person.active.of_dept(departments)
else
  people = people.titled_equal_or_higher_than(min_title)
end
people
Another alternative is the require a minimum title, which might be the simplest. I think NOT picking one would  be an oversight most of the times it would be blank. You could even have a title of `Everyone` with a very high value. This approach would eliminate the need for these changes.

goldenhub77:
@kgf was it decided it would be better to require a min_title instead of making logic changes?

kgf:
I am fine with requiring a minimum title.

goldenhub77:
@kgf Is there a reason why Rank 'Applicant' was removed from selection?

kgf:
Not that I can recall. However, that would not be a good choice for a default value, if that's what you were thinking. Recruit would be the lowest.
If you're in that area, I'd really like to move titles into a db setting. Most organizations will have different titles.

Thanks for your help.
Kevin

kgf:
Fixed by #692, by making min title a required field

Thanks @goldenhub77

PR #691, opened 2018-04-13
fixed bug #667

goldenhub77:


goldenhub77:
travis bug with spring preloader, i am resolving now

PR #692, opened 2018-04-13
fixed bug #667 and departments required for events

goldenhub77:


db/migrate/20180413193115_require_title_person.rb
```diff
@@ -0,0 +1,6 @@
+class RequireTitlePerson < ActiveRecord::Migration
+  def change
+    change_column_null :people, :title, false, "Recruit"
```
kgf, line 3: I worry some about putting this into the database. I would like to define the titles in the database so that other groups can define their own. They may not have any title of `recruit`. Perhaps if we changed this to `Unknown`, or something along those lines.

db/seeds.rb
```diff
@@ -60,14 +60,14 @@
    firstname: "Jane", lastname: "Doe", status: "Active", gender: "Female",
    start_date: 3.years.ago, department: mrc,
    nickname: "Janey", division1: "Division 1", division2: "Team 2",
-   icsid: "321", deployable: true
+   icsid: "321", deployable: true, title: "Recruit"
```
kgf: Could you please appease code climate, and use the single quotes here ? Thanks

spec/features/cert_displays_spec.rb
```diff
@@ -32,7 +32,7 @@
       expect(page).to have_link(person.fullname)
       expect(page).to have_link(cert.course.name)
       click_on person.fullname
-      expect(page).to have_content('CJ')
+      expect(page).to have_content('Cj')
```
kgf: ??

spec/models/person_spec.rb
```diff
@@ -233,6 +237,42 @@
     end
   end
 
+  describe 'calculate_title_order' do
+    let!(:director) {create :person, title: 'Director'}
```
kgf, line 241: There's no harm in having this in now, but it would need to come back out as we move titles into the database.

kgf:
This will need a rebase on `seeds.rb`. Small comments made, thanks for your efforts.

Kevin

db/migrate/20180413193115_require_title_person.rb
goldenhub77: @kgf As of right now, all logic determining events, people and there availabilities requires having a rank of 'Recruit' or higher or is will not calculate them as an availability. I can make an unknown rank as default but the default title_order number will have to be equal to 'Recruit' for everything to work properly.

spec/features/cert_displays_spec.rb
goldenhub77: @kgf  in order to properly format fullnames, I am using the capitalize method which uppercases the first letter and lower for the rest.

kgf: I don't know that changing the capitalization from what was entered is a good idea. What about names like MacDonald ?

goldenhub77: @kgf I can change this back if that is something that will be a problem

app/controllers/events_controller.rb
```diff
@@ -19,8 +19,7 @@ def archives
 
   def show
     @timecards = Timecard.active.overlapping_time(@event.start_time..@event.end_time)
-    @responses_and_assignments = ( @event.responses +
-       Assignment.active.for_time_span(@event.start_time..@event.end_time)).to_a
+    @responses_and_assignments = @event.responses + @event.assignments
```
kgf, line 22: I have a small problem with this change. Event.responses looks to availabilities the are in the time_span of the event. Assignments are a child (has_many) of an event. The trouble is that if I'm looking at Event A and I see Bob is available, I might try to assign him to Task 1. I would not know that he is already assigned to Task F in Event B, and thus is not actually available, unless I remove him from Task F.

goldenhub77: @kgf Ok, so you would like it to show all assignments within that events time frame?

db/migrate/20180413193115_require_title_person.rb
kgf: You implemented `Unknown` in some other places. Could we use that here ?

db/migrate/20180413193115_require_title_person.rb
```diff
@@ -0,0 +1,6 @@
...
+    change_column_null :people, :title_order, false, 30
```
kgf, line 4: And this to 100, if you agree with the comment above

db/migrate/20180413193159_require_min_title_events.rb
```diff
@@ -0,0 +1,5 @@
+class RequireMinTitleEvents < ActiveRecord::Migration
+  def change
+    change_column_null :events, :min_title, false, "Recruit"
```
kgf, line 3: I would prefer not to put the constraint in the DB, at least not until the titles are in the database and not hardcoded in the source. 
Also, wouldn't this prevent editing earlier events that may be in production and not have a min_title ? We may need to define one for those that don't currently have one.

db/migrate/20180413193115_require_title_person.rb
goldenhub77: Yes

db/migrate/20180413193115_require_title_person.rb
goldenhub77: Yes

db/migrate/20180413193159_require_min_title_events.rb
goldenhub77: This should also be 'Unknown', there has to be a constraint as it is a required entity and will crash the current application if there are values in the DB set to nil.

spec/features/events/event_page_spec.rb
```diff
@@ -79,12 +83,25 @@
           expect(page).to have_css('.event-labels[title="Assigned to THIS Event"]', text: '0')
           expect(page).to have_css('.event-labels[title="No Response"]', text: '0')
         end
+
+        visit event_path(different_event)
+
+        within("table#event-status tr##{@response_team.name.parameterize}-status") do
```
kgf, line 89: It's good having all these new specs. Thanks.

app/controllers/events_controller.rb
kgf: Yes, but that can be a different PR.
For those playing along at home, there is a use case for this. If you have someone showing as available for a time period, that will show on all events in that period. But you may have given them an assignment in Event B. If I'm doing assignments in Event A, I would think I could give them an assignment; they're available, but they're really not. I do not want to remove their availability from the list because I might think the assignment in Event A is more important, and pull them from Event B.

kgf:
Thanks for all your hard work on this @goldenhub77  !

kgf:
And thanks also to @farrahD , who started the work on Departments being required for events.

PR #693, opened 2018-04-23
Upgrade from rails 4 to rails 5

kgf:
This is a test to see how much work is involved...


kgf:
@goldenhub77 has a better PR in #695

PR #695, opened 2018-04-25
Rails upgrade to 5.2.0

goldenhub77:
Rails upgrade from 4.2.7 to 5.2.0 and Ruby environment tested in latest 2.5.1


goldenhub77:
FYI, there is a known bug with Reports - Calendar Chart that I have a fix and tests for that will be included in a separate PR and branch

goldenhub77:
Travis will not successfully run on ruby 2.5.1 until proper configurations can be put in place. This branch has been verified and tested on Ruby 2.5.1 but the main upgrade to Rails 5.2.0 will run on ruby 2.2+ so this is not a major issue at this time.

.ruby-version
```diff
@@ -1 +1 @@
-2.3.6
+2.5.1
```
kgf: If Travis is struggling to run on 2.5.1, and Rails 5 will run on 2.3.6 (2.2+) should we make this change. Should we make it to the newest version Travis supports ?

houhoulis: @kgf , why do you say Travis is struggling to run 2.5.1?  **_Ostensibly_**, they support it on Ubuntu and Mac builds.

kgf: See Sean's comment from yesterday.

pcai: @kgf so we're all on the same page: it looks like we're talking about https://travis-ci.org/ReadyResponder/ReadyResponder/builds/371111894 

Some googling turns this up: 
https://github.com/travis-ci/travis-ci/issues/8978

But im confused since it calls out rubygems 2.7.3, and the error is occuring on 2.7.6. The build is currently passing because the travis config of ruby 2.3x is overriding this `.ruby-version` file which declares 2.5.1. 

I can check the latest versions of bundler + rubygems @ mri 2.5.1 on travis if we want more information before continuing. But if things work fine in development and production on 2.5.1 its probably OK if travis runs using 2.3x, right?

pcai: on second thought, it seems better to not change ruby and rails version simultaneously unless we need to. after discussing with @kgf it seems best to have us on mri2.3.6 across all environments until this lands and stabilizes.

Issue #696, opened 2018-04-25
Unable to edit assignments due to requirement_id not being present


goldenhub77:
Files that will need to be modified: 
app/controllers/assignments_controller.rb
app/views/assignments/_form.html.erb
app/views/assignments/edit.html.erb

kgf:
Fixed by #698 
Thanks @goldenhub77  !

goldenhub77:
@kgf  It turns out there is a conflict when editing or creating a new assignment simply using just one _form so I created explicit forms for the edit and new actions in the controller until a better solution is determined.

kgf:
Thanks for your response. Could you elaborate on the conflict ?

kgf:
And I'm good with an expedient interim solution !

PR #698, opened 2018-04-26
bug fix for issue #696

goldenhub77:


kgf:
this is in reference to #696 regarding editing assignments

Issue #595, opened 2017-11-01
A task does not show the address


kgf:
Which makes it hard to use. You have to edit the task to show the address on the form.

Atul9:
@kgf I added a task & navigated to show page. I found that it shows the address i.e. the street name, city & state below the description. Can you please specify where is address not shown?

kgf:
I took a look, and this line is the issue
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/views/tasks/show.html.erb#L23
We had a call out this past week. People did not take the time to look up zip code, for example, which suppressed all of the address. Only entering the street would be a common use case.
Thanks for your help, @Atul9 !
Kevin

Atul9:
@kgf Do you suggest to have only @task.street.present? condition to display the address ?

kgf:
We should probably show whatever we have. Although having the town only would be unusual, and having the state only would be silly.

kgf:
@Atul9 Would you still be willing to help with this ?

AdManteza:
@kgf Can i take a stab at this one?

AdManteza:
@kgf I took a quick peek at the codebase. Couple of suggestions...

a.) Maybe we can require `street` and `city` for the `Task` model?
b.) After the `task` form is submitted, we can make a custom hook to somehow populate the `latitude/longitude`(~thru geolocation gem? cant remember exactly~ we already have it), using the street and city. If the user can provide a zip code(optional though), geolocation would be much more accurate. But at the very least we can generate the map properly.

Thoughts?

kgf:
I would prefer not to require the street or city. In many cases, the end users will be involved in only one town, meaning 90% of their tasks will be in that town. Requiring that field would force unnecessary data entry. I would prefer an approach of displaying whatever you have available.
Thanks for your help!
Kevin

kgf:
On point number two, there is a different issue for the Geo coding. We may have someone looking into that.

AdManteza:
@kgf 
Copy that. 
I'll just refactor the way we display the complete address in the view to account for missing data(e.g. street, zip code...). 
I'll spin up a PR for it if its ok with you.

kgf:
That's great, thanks !

kgf:
Closed by #703

PR #703, opened 2018-05-06
Issue#595 A Task does not show the address

AdManteza:
- [x] Use the view helper to render the address
- [x] Dry up and refactor the `tasks/show` page a bit

### DIFF
--- a/app/helpers/tasks_helper.rb
+++ b/app/helpers/tasks_helper.rb
@@ -20,12 +20,17 @@ module TasksHelper
   end
 
   def task_address
-    city = @task.city
-    city = @task.city + ', ' if @task.city.present?
+    comma = if @task.city.present? && @task.state.present?
+              ', '
+            else
+              ''
+            end
+
+    city_state = @task.city + comma + @task.state
 
     content_tag :div do
       concat content_tag(:p, @task.street)
-      concat content_tag(:p, city + @task.state + " #{@task.zipcode}")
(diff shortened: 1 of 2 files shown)


app/views/tasks/show.html.erb
```diff
@@ -21,18 +21,13 @@
   <p><b>Description:</b> <%= @task.description %></p>
 <% end %>
 
-<% if @task.street.present? and @task.city.present? and @task.state.present? and @task.zipcode.present? %>
-  <p>
-    <%= @task.street %><br>
-    <%= @task.city %>, <%= @task.state %> <%= @task.zipcode %>
-  </p>
-<% end %>
+<%= task_address %>
```
kgf, line 24: I like moving this to the helper

app/helpers/tasks_helper.rb
```diff
@@ -14,6 +14,21 @@ def task_priority_label(task)
...
   end
 
+  def has_latitude_longitude?
+    @task.latitude.present? and
+    @task.longitude.present?
+  end
+
+  def task_address
+    city = @task.city
+    city = @task.city + ', ' if @task.city.present?
```
kgf: Should we add the comma if the `@task.state` is missing ?

AdManteza: Good call!

AdManteza:
@kgf Ready for another look!

kgf:
This is regarding #595

kgf:
@AdManteza Thanks for your help !

PR #706, opened 2018-06-04
Certs <-- has and belongs to many --> Courses

masukomi:
After speaking with @kgf it was decided that there should be a HABTM association between courses and certs, because a course can convey multiple certs.

This PR adds that functionality.

On a related note: I don't really understand why a you associated a cert with courses instead of a course with certs. I changed the cert form to support multiple courses. It doesn't _feel_ quite right but i believe that it is, because there are multiple courses that could, for example, convey the same cert. For example the long First Responder / First Aid course includes CPR training but you could also have a short CPR course that conveyed the same CPR cert. 

That being said, it feels _wrong_ that you can't choose certs when creating / editing a course.

### DIFF
--- a/app/models/cert.rb
+++ b/app/models/cert.rb
@@ -2,13 +2,14 @@ class Cert < ActiveRecord::Base
   has_paper_trail
   include Loggable
 
-  attr_accessible :category, :person_id, :course_id, :expiration_date, :issued_date, :cert_number, :level,  :status, :certification, :comments
+  attr_accessible :category, :person_id, :courses, :expiration_date, :issued_date, :cert_number, :level,  :status, :certification, :comments
   belongs_to :person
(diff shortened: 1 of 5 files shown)


kgf:
Amazingly enough, I am open to renaming the models if it would make it easier for people to understand.

There is a person. A certification is that person completing some set of hurdles, such as taking a class and a test. An example of this would be a drivers license. They studied the book and then took the driving test. They are now licensed/certified to drive a car. This certification will expire. It is for one course.
A long first aid course with CPR gives two certs; First aid and CPR. It needs to work this way, because they expire on different timelines.
You aree quite correct that it does get complicated.

kgf:
@masukomi Would you be willing to freshen up this PR ? It did not pass the tests...

PR #708, opened 2018-06-07
set travis ruby to 2.3.3, same as .ruby-version

pcai:
Our goal is to have development and test using the same ruby version, so this just increments the patchlevel to match the value defined in https://github.com/ReadyResponder/ReadyResponder/blob/development/.ruby-version


kgf:
Sorry for the poor guidance. Could you please change *both* files to point to 2.3.3 ? ( That's assuming folks are not aware of a bad bug...)
That is what's currently running in production.
Thanks

kgf:
Thanks for your help @pcai

Issue #709, opened 2018-06-08
Unable to create an assignment


kgf:
Haven't seen the error log yet, but this is not good.

kgf:
Here is the report from the log when creating a new assignment. This only happens when creating a new assignment, not editing an existing one.
```
F, [2018-06-12T21:25:28.249454 #29907] FATAL -- :
ActionView::Template::Error (undefined method `event' for nil:NilClass):
    3:
    4:   <div class="form-inputs">
    5:     <%= f.association :person,
    6:                 collection: Person.active.where(:department => @assignment.requirement.event.departments).sort %>
    7:     <%= f.input :start_time, as: :string, label: "Start Time",
    8:                 input_html: { class: "datetimepicker", size: 11,
    9:                               value: format_datetime_value(f.object.start_time) } %>
  app/views/assignments/_form.html.erb:6:in `block in _app_views_assignments__form_html_erb__2021744206058970652_69872980698020'
  app/views/assignments/_form.html.erb:1:in `_app_views_assignments__form_html_erb__2021744206058970652_69872980698020'
  app/views/assignments/new.html.erb:6:in `_app_views_assignments_new_html_erb__3775128160103050522_69872987036680'
```

kgf:
This is related to #696.
@goldenhub77 do you recall that work ?

kgf:
I will create a topic branch based off of v2.10.0, it will be `rel2.10.1`
Any PRs should be against that and also the head of development.
Thanks

PR #714, opened 2018-06-14
fixed create an assignment bug #709

goldenhub77:


app/controllers/assignments_controller.rb
```diff
@@ -15,6 +15,7 @@ def show
 
   def new
     @assignment = Assignment.new
+    @assignment.requirement = @requirement
```
kgf, line 18: Isn't this covered by line 5, the before hook ?

kgf:
@goldenhub77 Thanks for your assistance. Only one comment.

Could you also PR this against v2.10.0 branch ? That's what's in production now

app/controllers/assignments_controller.rb
goldenhub77: @kgf This is a necessary change. I had to change the forms and assignment needs to have a requirement set before it gets to the view or you will get no method for null errors while its loading the form.

kgf: Ah, I didn't parse this correctly the first time. The before hook **is** defining the requirement, and here you're making it the parent to the new assignment.

kgf:
Thanks @goldenhub77

PR #715, opened 2018-06-14
bug fix for bug #709

goldenhub77:


kgf:
This is pulling in all the Rails 5 stuff. I suspect the head of development on Sean's fork is even with the HEAD the main repo's development.

I would appreciate a PR **based** on the v2.10.0 branch and also targeting that branch, with only these changes. We're still testing the major changes in Rails 5.2 upgrade

PR #720, opened 2018-07-05
Added Rails version to migration definitions

kgf:
See https://blog.bigbinary.com/2016/03/01/migrations-are-versioned-in-rails-5.html


db/schema.rb
```diff
@@ -324,6 +324,7 @@
     t.string "event_statuses"
     t.string "event_assigned"
     t.string "id_code"
+    t.text "individuals"
```
pcai, line 327: which is correct?

PR #722, opened 2018-07-05
Added default subject message back in

### LINKED ISSUE #719: New notifications no longer have default message.
The default used to include details about the event, but now it stops at `Please provide availability for `

kgf:
Fixes #719


pcai (approved):
i was going to say this feels like a model concern (e.g., one that could be an `after_initialize` in the Notification model), but on second thought it seems closer to a presentational one. 👍

Issue #710, opened 2018-06-08
When looking at event#show, availabilities with a status of canceled are shown


eroc7886:

kgf:
Cancelled availabilities should only be shown on the person#show page.

aadeshere1:
@kgf  Could you check the PR ?

PR #726, opened 2018-10-02
select only the availabilities whose status is available

aadeshere1:
closes issue #710

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -19,7 +19,7 @@ class EventsController < ApplicationController
 
   def show
     @timecards = Timecard.active.overlapping_time(@event.start_time..@event.end_time)
-    @responses_and_assignments = @event.active_responses + @event.assignments.active
+    @responses_and_assignments = @event.responses + @event.assignments.active
     @page_title = @event.title
     @last_editor = last_editor(@event)
   end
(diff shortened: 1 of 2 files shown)


aadeshere1:
@kgf seems this test is failng from the `development` branch.

kgf:
Thanks for your help!
The test suite is correct, because you aren't showing unavailable folks. If you change your select to this scope, I think all will be well.
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/models/availability.rb#L17

aadeshere1:
@kgf I didn't understand the part about changing select to the scope you referred to. Instead i changed the `available` to `cancelled`

kgf:
Thanks for your help.
Scopes are basically a shortcut for what you did in defining a new method. If you change the line in the controller to `event.responses.active`, that will use the scope I referenced earlier. You won't need the change to the model. 
The scope also handles the case of hiding availabilities if people becoming inactive. Scopes typically have better performance.

kgf:
Here is the guide
https://guides.rubyonrails.org/active_record_querying.html#scopes

aadeshere1:
Thanks @kgf. Updated

kgf:
Thanks @aadeshere1 
Since you're in this part of the code, would you be willing to add some tests to ensure cancelled doesn't appear again?

aadeshere1:
@kgf Sure. I'll add the tests too.

PR #732, opened 2018-10-02
Remove outdated changelog

greysteil:
It looks like this changelog is now very outdated (the last entry was 2013), so should probably be removed.

I'm also interested in whether the tests pass with this cosmetic change, to see whether the failures on #729, #730 and #731 are genuine.


kgf:
That's a clever tactic. I agree, this is probably not related !

kgf:
This is no longer needed, we got the test suite going again.

greysteil:
Awesome - sorry, was midnight here and I needed some rest. Thanks for picking up!

kgf:
No problem. And I think I'll update the changelog from my comments in the releases section. The test suite fix was all thanks to @buitranquyet

Issue #598, opened 2017-11-05
Events archive date columns sort by ASCII rather than by date


VanAuken:

vsrnth:
Is there some sample data to reproduce this bug?

PR #741, opened 2018-10-05
events archive sort by date

Erin-Christensen:
added data-order attribute to the start-time and end-time <td> tags. This allows these columns to be sorted by the unformatted timestamps.
Fixes issue #598

### DIFF
--- a/app/views/events/archives.html.erb
+++ b/app/views/events/archives.html.erb
@@ -18,10 +18,10 @@
       <tr class='<%= column_color_class(e) %>'>
         <td><%= link_to e.title, e %></td>
         <td><%= e.location %></td>
-        <td><time>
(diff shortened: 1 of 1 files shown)


kgf:
Thanks for your help !
I'm not familiar with this approach. We don't need to add hidden columns for the sorting, like this ?
https://github.com/ReadyResponder/ReadyResponder/blob/development/app/views/people/index.html.erb#L17
Kevin

Erin-Christensen:
I found it in the data tables manual [here](https://datatables.net/manual/data/#DOM).
I tested it with events scheduled for each day of the week, and it's sorting by the date now, where it was sorting alphabetical by day of the week before. So it seems the hidden columns are unnecessary with this approach.

kgf:
Thanks @Erin-Christensen for your help !

Issue #712, opened 2018-06-14
Rank should not be required for Prospect on New Person


VanAuken:
When making a New Person, people are often entered as Prospect or Applicant.  Rank is required in order to save, and currently we are using 'unknown' to meet requirements.

Accurate logic would not require a rank for 'prospect' or 'applicant'.  A rank is only needed for people who have joined.  This would also require a blank or null choice for rank.

yasirazgar:
I can look into this.

houhoulis:
@VanAuken ,

Considering a person might have a Status such as "Applicant", I approve of removing the requirement that a "Rank" (title) be entered for them.  However, I'm not certain, but this issue might be requesting to require a "Rank" for certain statuses but not others.  I'm not in favor of hard-coding that level of logic into the app; could we just make "Rank" optional for all people?

I'm not in favor of hard-coding 'Rank required for Statuses X, Y, and Z' logic because I'm not sure that all organizations will have the same idea about which Statuses should require a Rank.  Until we have more configurability built into the app, I'd rather avoid such detailed restrictions.

Can we make "Rank" optional for all people?

kgf:
I agree with Chris, both about not trying to hard code a matrix, and not requiring title/rank for anyone at this time.

karmenlu:
I would like to work on/claim this issue. @kgf

kgf:
Thanks, @klu17

PR #743, opened 2018-10-18
712 - Don't require rank. [edited: was "Require rank only for active person"]

yasirazgar:
Fixes, https://github.com/ReadyResponder/ReadyResponder/issues/712

When creating or editing a person, rank/title should be required only for person with active status.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -35,7 +35,6 @@ class Person < ApplicationRecord
   validates_numericality_of  :height, :weight, allow_nil: true, allow_blank: true
   validates_presence_of :division2, unless: -> { division1.blank? }
   validates_presence_of :division1, unless: -> { division2.blank? }
-  validates_presence_of :title, if: -> { active? }
   validates_chronology :start_date, :end_date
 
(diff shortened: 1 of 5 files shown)


houhoulis:
Hi, @yasirazgar !  Thanks for helping!  Sorry for the delay reviewing your PR.

Could you add a migration to implement the database change?

I also have a question for @VanAuken : #712 requests that "Rank" not be required for statuses "Prospect" or "Applicant".  This PR makes a slightly different change: it _only_ requires "Rank" if the status is "Active".  Should "Rank" be required for the other statuses `'Leave of Absence'`, `'Inactive'`, and `'Declined'`?

houhoulis:
@yasirazgar , it looks as though we can remove the requirement & validation on "Rank" (title) altogether, rather than requiring it conditionally based on Status.  See https://github.com/ReadyResponder/ReadyResponder/issues/712#issuecomment-433659326

yasirazgar:
@houhoulis, removed validation for person#title altogether.

houhoulis (approved):
thanks!  👍

Issue #746, opened 2018-11-07
Upcoming events does not honor event status


kgf:
Currently, when texting `upcoming` into Ready Responder, people get really old events.
This line needs to filter events to only upcoming. There is already a scope in the event model to handle this.

https://github.com/ReadyResponder/ReadyResponder/blob/development/app/models/person.rb#L232

kgf:
Thanks to @lilsweetcaligula and @houhoulis for resolving this. I can confirm it works in production.

Thanks!

PR #747, opened 2018-12-02
Make Person#upcoming_events honor the event status

lilsweetcaligula:
#746 

Awaiting any requests for changes that might come up.

P. S. I'm personally doubting if the `:upcoming` trait that I have defined is really needed, but felt it was sensible to include it. What do you people think?

### DIFF
--- a/spec/models/person_spec.rb
+++ b/spec/models/person_spec.rb
@@ -293,22 +293,29 @@ RSpec.describe Person do
   describe 'upcoming_events' do
     let(:department) { create :department }
     let(:person) { create :person, department: department }
-    let(:e1) { create :event, departments: [department] }
-    let(:e2) { create :event, departments: [department] }
-    let(:e3) { create :event, departments: [department] }
-    let(:e4) { create :event, departments: [department] }
-    let(:e5) { create :event, departments: [department] }
-    let(:e6) { create :event, departments: [department] }
-    let(:e7) { create :event, departments: [department] }
-    let(:e8) { create :event, departments: [department] }
-    let(:e9) { create :event, departments: [department] }
-    let(:e10) { create :event, departments: [department] }
+    let(:e1) { create :event, :upcoming, departments: [department] }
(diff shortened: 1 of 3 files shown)


houhoulis:
This looks very good to me.  I agree with your assessment, the `:upcoming` trait seems useful though not essential.

spec/models/person_spec.rb
```diff
@@ -293,23 +293,30 @@
...
     context 'with settings' do
       it 'returns count from setting' do
         expect(person.upcoming_events).to eq [e6, e7, e8, e9, e10]
       end
+
+      it 'returns only the upcoming events' do
+        e7.update!(status: 'Closed')
+        e9.update!(start_time: 2.days.ago, end_time: 1.day.ago)
+
+        expect(person.upcoming_events).to eq [e6, e8, e10]
```
houhoulis, line 318: The `person.upcoming_events` query, as far as I can tell, is not guaranteed to return events in a particular order.  When you're testing a result which should be acceptable in any order, I would recommend using `match_array` or `contain_exactly` instead of `eq`.  That way, if the database happens to return events in a different order, the test will still pass.

All the other tests in this `'upcoming_events'` block already use `eq`, though, so I'm fine merging as is.

houhoulis (approved):
Thanks @lilsweetcaligula !  👍

spec/models/person_spec.rb
lilsweetcaligula: @houhoulis 

Thank you! Would you like me to create an issue, go through this particular spec and change `eq` to `match_array`, where appropriate?

Your suggestion of using the `match_array` additionally has the potential of making the intention of the unimportant order in the results more clear as well.

houhoulis: Thank you @lilsweetcaligula for the offer!  I saw this comment immediately after I made #748 .  Next time I'll take you up on it!

PR #748, opened 2018-12-02
Change `eq` to `contain_exactly` in specs where...

houhoulis:
... we don't care about the order of the results.

I didn't find other specs using `eq` with an array that didn't care about order.

cc @lilsweetcaligula


lilsweetcaligula:
:+1: Cool, thanks!

kgf:
Thanks @houhoulis  !

PR #753, opened 2019-07-03
Containerizing it

masukomi:
added all the docker files


masukomi:
Travis is failing because of a bunch of these:

```
      Selenium::WebDriver::Error::UnknownError:
        unknown error: Chrome failed to start: exited abnormally
          (unknown error: DevToolsActivePort file doesn't exist)
          (The process started from chrome location /usr/bin/google-chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
```

locally i did a `bundle update` and that resolved the issue. I've checked in the updated `Gemfile.lock`  but that didn't seem to help Travis. 

when run in the container you see this

```
Selenium::WebDriver::Error::WebDriverError:
        unable to connect to chromedriver 127.0.0.1:9515
```

With the updated Gems i now have a pile of these at the start of the run:

```
DEPRECATION WARNING: Static attributes will be removed in FactoryBot 5.0. Please use dynamic
attributes instead by wrapping the attribute value in a block:

comments { "" }

To automatically update from static attributes to dynamic ones,
install rubocop-rspec and run:

rubocop \
  --require rubocop-rspec \
  --only FactoryBot/AttributeDefinedStatically \
  --auto-correct

 (called from block (2 levels) in <top (required)> at /app/spec/factories/vendor.rb:9)

```

that should be broken out into its own ticket after this one is merged. 


As an aside, running tests within the docker container is literally 10x slower. Dunno why.

.ruby-version
```diff
@@ -1 +1 @@
-2.3.3
+2.6.0
```
kgf, line 1: @masukomi This version seems to only be in preview for rbenv ? Would 2.5.3 be new enough ? ( I wonder if the 2.6 preview has additional debug code, which is what causes the slowness of the tests.

kgf:
@masukomi Were you saying you wanted to close this ?

masukomi:
Closing in favor of #768

PR #754, opened 2019-10-14
Ignore Cancelled Tasks

### LINKED ISSUE #751: Cancelled Assignments still count for the colored summary boxes.
Cancelled assignments should not affect the count.
These are the boxes I am referring to....
Screen Shot 2019-05-26 at 8 39 18 PM

yez:
Closes #751

Only have tasks on an assignment that are not canceled.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -10,7 +10,11 @@ class Assignment < ApplicationRecord
 
   validates_presence_of :person, :start_time, :end_time
 
-  STATUS_CHOICES = [ 'New', 'Active', 'Cancelled' ]
+  STATUS_CHOICES = [
+    STATUS_NEW = 'New',
+    STATUS_ACTIVE = 'Active',
+    STATUS_CANCELLED = 'Cancelled'
+  ]
 
   scope :active, -> { where( status: ["New", "Active"] ) }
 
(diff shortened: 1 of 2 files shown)


kgf:
Howdy, @yez Thanks for your help ! 
I see you converted the status choices from an array of strings to a hash. Will the form still work with this change ?
https://github.com/ReadyResponder/ReadyResponder/blob/2a1d1bb99c532923a8239f3b9c4609a37b70e298/app/views/assignments/_edit_form.html.erb#L13

yez:
@kgf It is actually still an Array. I just added assignment within the Array to expose a set of new constants that can be used as well. It is a little tricky but the contract remains the same. 

```ruby
2.3.3 :001 > Assignment::STATUS_CHOICES
 => ["New", "Active", "Cancelled"]
2.3.3 :002 > Assignment::STATUS_CHOICES.class
 => Array
```

kgf:
Thanks @yez that's quite helpful. Could we interest you in another ?

yez:
@kgf I probably have time for more work, what did you have in mind?

kgf:
Well, there are some defined issues already. I would especially like to support email notifications, and also tagging on both people and items. 
But if you have something that you want to explore ( i.e. front end dev ), I could probably create an issue to match. We use this app all the time, and I have a lot of ideas....

Thanks for your interest

yez:
@kgf I'll give those issues a look in my free time!

PR #759, opened 2019-10-31
Added a test to avoid method on null

kgf:
Cuz that never turns out well.

### DIFF
--- a/app/views/analytics/system_activity_logs.html.erb
+++ b/app/views/analytics/system_activity_logs.html.erb
@@ -15,7 +15,7 @@
   <tbody>
     <% @logs.each do |log| %>
       <tr>
-        <td><%= log.user.fullname %></td>
+        <td><%= log.user.present? ? log.user.fullname : "Unknown User" %></td>
         <td><%= log.category %></td>
         <td><%= log.message %></td>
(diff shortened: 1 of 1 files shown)


app/views/analytics/system_activity_logs.html.erb
```diff
@@ -15,7 +15,7 @@
   <tbody>
     <% @logs.each do |log| %>
       <tr>
-        <td><%= log.user.fullname %></td>
+        <td><%= log.user.present? ? log.user.fullname : "Unknown User" %></td>
```
gasparyn: also
```
<td><%= log&.user&.fullname || "Unknown User" %></td>
```

kgf: Well, `log` can't be null, so `<%= log.user&.fullname || "Unknown User" %>` would also be acceptable ?

gasparyn: even better. 👍

gasparyn:
looks like we need to rebase with development too

gasparyn (changes requested):
rebase needed

gasparyn (approved):

PR #761, opened 2019-10-31
Added tagging to the big three models

kgf:
Person, event and item.


gasparyn:
can we add 3 small unit tests for each model that is now :taggable?
Mostly to make sure that they are actually acting as taggable

houhoulis:
I agree with @gasparyn. Simple tests would suffice. I see the benefits as
1. ensuring taggable is configured & working.
1. documenting the use of taggable.

